PR #49, opened 2013-02-20
Add next_marker method (#48

methane:


methane:
I don't have motivation to add APIs increase complexity and maintenance cost anymore.

PR #100, opened 2014-05-22
added distinguish_tuple argument to Packer

faerot:
This will make precise python types serialization possible.

### DIFF
--- a/msgpack/_packer.pyx
+++ b/msgpack/_packer.pyx
@@ -56,10 +56,13 @@ cdef class Packer(object):
         Convert unicode to bytes with this encoding. (default: 'utf-8')
     :param str unicode_errors:
         Error handler for encoding unicode. (default: 'strict')
-    :param bool distinguish_tuple:
-        If set to true, tuples will not be serialized as lists
-        and will be treated as unsupported type. This is useful when trying
-        to implement accurate serialization for python types.
+    :param bool precise_mode:
+        If set to true, types will be checked to be exact. Derived classes
+        from serializeable types will not be serialized and will be
+        treated as unsupported type and forwarded to default.
+        Additionally tuples will not be serialized as lists.
+        This is useful when trying to implement accurate serialization 
+        for python types.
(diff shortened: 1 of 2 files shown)


methane:
simplejson has 'tuple_as_array' option from 2.2.0.  I'll add it.
https://simplejson.readthedocs.org/en/latest/#simplejson.JSONEncoder

msgpack/_packer.pyx
```diff
@@ -56,6 +56,13 @@ cdef class Packer(object):
         Convert unicode to bytes with this encoding. (default: 'utf-8')
     :param str unicode_errors:
         Error handler for encoding unicode. (default: 'strict')
+    :param bool precise_mode:
+        If set to true, types will be checked to be exact. Derived classes
+        from serializeable types will not be serialized and will be
+        treated as unsupported type and forwarded to default.
```
methane: I don't like `precise_mode` name since it doesn't mention about types.
What about `strict_types`?

msgpack/_packer.pyx
```diff
@@ -184,7 +195,7 @@ cdef class Packer(object):
                         if ret != 0: break
                         ret = self._pack(v, nest_limit-1)
                         if ret != 0: break
-            elif PyDict_Check(o):
+            elif not precise and PyDict_Check(o):
```
methane: Did you miss `PyDict_CheckExact(o) if precise else`?

msgpack/fallback.py
```diff
@@ -507,18 +516,30 @@ def __init__(self, default=None, encoding='utf-8', unicode_errors='strict',
                 raise TypeError("default must be callable")
         self._default = default
 
-    def _pack(self, obj, nest_limit=DEFAULT_RECURSE_LIMIT, isinstance=isinstance):
+    def _check_precise(obj, t, type=type, tuple=tuple):
+        if type(t) is tuple:
```
methane: This function should be moved to out of class since it's not method.

PR #123, opened 2015-01-23
tests: add pytest 2.3 compatibility

ktdreyer:
Adjust the skipif conditional to use the older pytest 2.3 syntax.

(This allows the tests to pass with the system pytest package on RHEL 7.0, since RHEL 7.0 ships pytest 2.3.5. See downstream bug https://bugzilla.redhat.com/1182808)

### DIFF
--- a/test/test_unpack.py
+++ b/test/test_unpack.py
@@ -16,7 +16,7 @@ def test_unpack_array_header_from_file():
         unpacker.unpack()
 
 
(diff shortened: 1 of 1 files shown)


methane:
Thanks

ktdreyer:
Thanks @dmick :) and thanks @methane for reviewing, merging, and releasing this in 0.4.5!

Issue #133, opened 2015-03-16
Large python integer can't be handled with hooks


pramukta:
This makes it impossible to work around an issue such as #71.  Here's a test that replicates the problem, based on the [example in the docs](https://github.com/msgpack/msgpack-python/blob/master/README.rst#packingunpacking-of-custom-data-type):

```python
 def test_overriding_hooks():
     def default(obj):
         if isinstance(obj, long):
             return {"__type__": "long", "__data__": str(obj)}
         else:
             return obj

     obj = {"testval": long(1823746192837461928374619)}
     refobj = {"testval": default(obj["testval"])}
     refout = msgpack.packb(refobj)
     assert isinstance(refout, str)
     testout = msgpack.packb(obj, default=default)

     assert refout == testout
```

The issue looks to be that the logic to run the default handler is executed at the end of the while block, after it has already tried to handle a bunch of types.  If any of those types match, the provided handler is not called.  In this case, the built-in handler errors out due to the number being too large ([which is expected](https://github.com/msgpack/msgpack/blob/master/spec.md#types-limitation)) and there is no way (that I can see) to extend the serialization to handle the case for a particular application.

I'm still trying to figure out a workaround/patch myself, but if anyone else has thoughts, it would be great to hear.   Other than preparing the data so that it doesn't have long integers (since I'm not in control of that).  Thanks for looking.

pramukta:
Updating the title of this issue to be consistent with the pull request that addresses it (#135) so that there is a greater chance that it gets assessed / integrated.

pramukta:
This issue was addressed via merging #135

PR #134, opened 2015-03-17
Make sure the provided default function is always called when packing

pramukta:
This is to address issue #133.  All the tests pass with both the cython and fallback versions.  (I think, am not 100% clear on the cython workflow).  Let me know if there is anything wrong or that should be better or if I'm off base about anything.

### DIFF
--- a/msgpack/_packer.pyx
+++ b/msgpack/_packer.pyx
@@ -124,101 +124,98 @@ cdef class Packer(object):
 
         if nest_limit < 0:
             raise PackValueError("recursion limit exceeded.")
+        if self._default is not None:
+            o = self._default(o)
 
-        while True:
-            if o is None:
-                ret = msgpack_pack_nil(&self.pk)
-            elif isinstance(o, bool):
+        if o is None:
+            ret = msgpack_pack_nil(&self.pk)
+        elif isinstance(o, bool):
                 if o:
                     ret = msgpack_pack_true(&self.pk)
                 else:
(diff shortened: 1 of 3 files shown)


pramukta:
Based on the CI it looks like I didn't address python 3.  Looking into it.

pramukta:
Ok, looks like it's good to go.  Let me know if there's anything else I should do.

methane:
I'm sorry, but I will not merge this breaking API change.
`default` keyword argument comes from [here](https://docs.python.org/3.3/library/json.html#json.dump).

Please consider about existing code like below:

```python
def default(obj):
    if isinstance(obj, datetime):
        return obj.timestamp()
    raise TypeError("Unsupported type %s" % (type(obj),))
```

methane:
I think it's OK to call `default` only when integer overflow.

I'll drop Python 2.6 support when msgpack-python 0.5.
It is easy to check overflow with this API.
https://docs.python.org/2.7/c-api/long.html#c.PyLong_AsLongLongAndOverflow

pramukta:
Ah I see, good to know where that API came from.  I had assumed that it was supposed to be used the way it was documented [here](https://github.com/msgpack/msgpack-python/blob/master/README.rst#packingunpacking-of-custom-data-type).  It might be a good idea to update those docs to match the test you posted.  In the mean time, I'll see if I can implement it closer to the way you've suggested.

pramukta:
Closing in favor of the alternate implementation #135

PR #135, opened 2015-03-17
Default function on int overflow

pramukta:
Here's an implementation in the same spirit as the ones suggested in comments to pull request #134, also addressing issue #133, in the specific case of long integers that overflow.

### DIFF
--- a/msgpack/_packer.pyx
+++ b/msgpack/_packer.pyx
@@ -136,12 +136,20 @@ cdef class Packer(object):
             elif PyLong_Check(o):
                 # PyInt_Check(long) is True for Python 3.
                 # Sow we should test long before int.
-                if o > 0:
-                    ullval = o
-                    ret = msgpack_pack_unsigned_long_long(&self.pk, ullval)
-                else:
-                    llval = o
-                    ret = msgpack_pack_long_long(&self.pk, llval)
+                try:
+                    if o > 0:
+                        ullval = o
+                        ret = msgpack_pack_unsigned_long_long(&self.pk, ullval)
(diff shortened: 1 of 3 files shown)


pramukta:
This one should avoid the breaking API change.  Thoughts?

pramukta:
Any progress on getting this merged?  If there is anything else needed on it let me know, I'd be happy to get it done.

pramukta:
Any project collaborators want to weigh in on whether this will be merged or if there are additional things to do?  I know everyone is very busy and the issue this corrects may seem like a small thing (though the issue does technically make msgpack and JSON incompatible in this implementation), but this fix is very useful to me and I'm happy to do whatever work is necessary to push this along.

Needless to say, a great way to kill people's motivation to contribute to an open source project is to provide no responses or feedback for months at a time after someone has tried to make an honest effort.  I'm sure that's not the intent here and that the contributors are just busy and distracted.  Just hoping this message will get someone's attention so I can avoid having to maintain a separate branch myself at the expense of potential community benefit.

methane:
Sorry for long no response.
I'm very hard on private. (I have 1 year old daughter and my wife is PMS and BPD.)
I have very limited time and energy.

I hope someone triage and review requests.
But msgpack-python is very complicated already compared with other languages' msgpack library.
So I want collaborator is very careful against making msgpack-python complicated.
If come true, remove one feature before merging one new feature....

pramukta:
Thanks!  I definitely appreciate the hard work and the push to keep things simple.  Thanks for the great library.

PR #141, opened 2015-06-02
Support for serializing to Str and Bin from Python 2.X

ambrusc:
When setting `use_bin_type=True` in Python 2.X, we lose the ability to serialize strings as Str (Raw); instead, all `str` objects are serialized as Bin.

You probably already know that in Python 2.X, `bytes == str` so the serializer has no way to distinguish between encoded and un-encoded data passed as a string. Unfortunately I found no mechanism to use `default` to get the desired behavior (`str` => Str and `[something]` => Bin). It was either "everything is Str" or "everything is Bin".

Proposed change:
* backward-compatible behavior when `use_bin_types == False` (everything => Str (Raw))
* add serialization of `bytearray` to Bin or Str as determined by `use_bin_types`
* enables (`str` => Str, `bytearray` => Bin) in Python 2.X
* existing Python 3.X behavior unchanged: (`str` => Str, `bytes` => Bin)

This is my first pull request to an open source project so please go easy! I'd really like to help support the community and I'm very excited to use msgpack. It's already sped up one of my protocols by 1000x and I'd love to keep helping to improve it.

Thoughts?

### DIFF
--- a/msgpack/_packer.pyx
+++ b/msgpack/_packer.pyx
@@ -2,6 +2,7 @@
 #cython: embedsignature=True
 
 from cpython cimport *
+from cpython.version cimport PY_MAJOR_VERSION
 from libc.stdlib cimport *
 from libc.string cimport *
 from libc.limits cimport *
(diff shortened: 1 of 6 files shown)


ambrusc:
Thanks for the help on twitter, @methane. Canonicalizing data before passing to the encoder is a much better way to get the right behavior so I'm closing this PR.

Issue #144, opened 2015-08-25
[Feature Request] Ability to reset unpacker


hendrikmuhs:
Hi,

it would be nice to be able to reset the state of an Unpacker object. If I get it right it is just a matter of resetting self.buf_{head, tail} to 0. Correct?

PR #145, opened 2015-08-26
Unpacker reset

hendrik-cliqz:
see #144

### DIFF
--- a/msgpack/_unpacker.pyx
+++ b/msgpack/_unpacker.pyx
@@ -458,6 +458,11 @@ cdef class Unpacker(object):
     def __next__(self):
         return self._unpack(unpack_construct, None, 1)
 
+    def reset(self):
+        """Clear internal buffer."""
+        self.buf_head = 0
+        self.buf_tail = 0
+
     # for debug.
     #def _buf(self):
     #    return PyString_FromStringAndSize(self.buf, self.buf_tail)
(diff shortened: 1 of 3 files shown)


methane:
Unpacker takes file-like object in constructor.
This PR doesn't care about it.

hendrik-cliqz:
You are right. As I do not use it with a file-like object I simply overlooked this usecase.

But I am wondering about the way to handle it: It seems to me that reset does not make sense if it reads from a file-like object. So I would throw as in feed(). Correct?

methane:
What is motivation of reset?
Is it really unrelated to file-like object usage?

hendrik-cliqz:
sorry for the late reply. I am using the reset method in a RPC library - so essentially sending msgpack over the network:

https://github.com/hendrik-cliqz/mprpc/blob/master/mprpc/client.pyx#L137

The unpacker object in there is reused, to ensure it is ok to use it, I reset it. I want to avoid re-creating the object for every call due to perf reasons.

methane:
OK.  So, why `unpacker.feed(file=None)` doesn't make sense?

hendrik-cliqz:
sorry, I do not get the question. "unpacker.feed(file=None)" isn't a valid call, is it?

Actually I do not need this PR anymore, so I suggest to close it.

methane:
OK.  Thanks anyway.

PR #161, opened 2015-11-18
Support packing memoryview objects

jfolz:
I am working on a high-throughput scenario where I need to avoid copying memory wherever possible. I frequently create buffers from existing objects and sometimes need to hand them to msgpack.
Unfortunately I could not find a way to pass buffers to the current version of msgpack without copying.
So I wrote this small patch to support packing arbitrary buffer objects to binary data. It works well for my use case.

Things I'm not so sure about:
* Pure Python fallback is not implemented. Its structure is different from the Cython code and I'm not sure what to do exactly.
* I'm not an expert in using this library. Does this interfere with any other functionality?
* Information is lost in the conversion, e.g., shape of Numpy arrays. Since the lib just "swallows" everything that is a buffer. A user might expect to get the same type out again, but gets bytes instead.
* Strided data is still copied.
* No test cases yet. Might needs some more type checks.

### DIFF
--- a/msgpack/_packer.pyx
+++ b/msgpack/_packer.pyx
@@ -10,6 +10,11 @@ from msgpack.exceptions import PackValueError
 from msgpack import ExtType
 
 
+cdef extern from "Python.h":
+
+    int PyMemoryView_Check(object obj)
+
+
 cdef extern from "pack.h":
     struct msgpack_packer:
         char* buf
@@ -132,6 +137,7 @@ cdef class Packer(object):
         cdef size_t L
         cdef int default_used = 0
         cdef bint strict_types = self.strict_types
+        cdef Py_buffer view
 
         if nest_limit < 0:
             raise PackValueError("recursion limit exceeded.")
@@ -231,6 +237,17 @@ cdef class Packer(object):
                     for v in o:
                         ret = self._pack(v, nest_limit-1)
                         if ret != 0: break
+            elif PyMemoryView_Check(o):
+                if PyObject_GetBuffer(o, &view, PyBUF_SIMPLE) != 0:
+                    raise ValueError("could not get buffer for memoryview")
+                L = view.len
+                if L > (2**32)-1:
+                    PyBuffer_Release(&view);
+                    raise ValueError("memoryview is too large")
+                ret = msgpack_pack_bin(&self.pk, L)
+                if ret == 0:
+                    ret = msgpack_pack_raw_body(&self.pk, <char*>view.buf, L)
+                PyBuffer_Release(&view);
             elif not default_used and self._default:
                 o = self._default(o)
                 default_used = 1
--- a/msgpack/fallback.py
+++ b/msgpack/fallback.py
@@ -36,6 +36,8 @@ if hasattr(sys, 'pypy_version_info'):
             else:
                 self.builder = StringBuilder()
         def write(self, s):
+            if isinstance(s, memoryview):
+                s = s.tobytes()
             self.builder.append(s)
         def getvalue(self):
             return self.builder.build()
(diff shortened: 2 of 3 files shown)


jfolz:
I changed the order of evaluation so that the default handler is used before checking for buffer support. Python 3 tests are now passing.
To use ExtTypes and buffers at the same time, a default function simply has to return the buffer unmodified.

thedrow:
What about the pure python implementation?

bwesterb:
Yeah, you should have added tests for the new behaviour.

jfolz:
You're both right of course. Those points are on my list.
Before I put more work into this I just wanted to discuss if this is desired behavior. As mentioned, just accepting anything regardless of type could be confusing. E.g., a user encodes ndarrays and gets raw bytes back. To make users aware of what is going on it might make sense to require a default function to encode buffers.
Also, buffers can specify dtype and strides. That is basically the way ndarrays are described everywhere. AFAIK this is not directly supported by the spec, but there may be a preferred way of doing this that I'm not aware of?
Those are basically the points that I would like to discuss here. I'm not very keen on keeping my fork alive forever, but I still need to performance I get from it.

methane:
I'm afraid about accept anything implements buffer interface.
How about accept only `memoryview` object?
You can convert your data implements buffer interface into memoryview instead of bytes.  It avoids copy

jfolz:
That could work. Creating memoryviews is very cheap and this extra step makes the process much more explicit than just accepting any old buffer. It would still use the buffer interface to access the data referenced by the memoryview though.

jfolz:
Been working on this today. Expected py2.6 to fail (no memoryview) and PyPy failed because StringBuilder doesn't accept memoryviews (should work now).
Haven't been successful accessing PY_MAJOR_VERSION/PY_MINOR_VERSION at Cython compile time to place compiler directives. Don't really want to check for Python version every time an object is packed.

jfolz:
I got the version switch working (now in a separate header) and it's all green on 2.7+ and PyPy, however 2.6 is failing still.
From the logs it looks it builds and installs fine, but cannot import the extension modules when tests are run.

thedrow:
@jfolz See http://stackoverflow.com/questions/28299202/how-to-conditionally-declare-code-according-to-python-version-in-cython/28316509#28316509 regarding getting the Python version at compile time.

methane:
I'm OK to drop Python 2.6 support.

jfolz:
@thedrow Unfortunately that doesn't seem to work, at least not in my version of Cython (0.20.1; Ubuntu 14.04).

@methane That would certainly make things easier ;)

thedrow:
0.20 is pretty old.
Why won't you upgrade to the latest?

jfolz:
I got it working for py26 using `buffer` instead of `memoryview` objects as they're available. Unfortunately even though the new buffer interface itself was backported, no support to export was added to the builtin types.
I could try to use the old-style buffers on py26, but this is a dead end anyway. My 'solution' is to just do nothing for py26.
Using py27+ you can now create memoryviews from your objects and have msgpack pack them.

@thedrow That's what I got from the repository, so I suppose it's what a lot of people will be using. The workaround really isn't that bad, it being one tiny header file. Plus this way I can use the proper preprocessor and don't need to rely on Cython's somewhat limited and awkward version.

thedrow:
The build always uses the latest Cython. So it's probably better to use that when developing.

jfolz:
@methane What's your opinion on the current state of the patch?

methane:
@jfolz I've released 0.4.7 and removed Python 2.6 support in current master.
Rebase (or merge) master and remove ugly code for support Python 2.6.

jfolz:
@methane I removed py26 stuff and rebased. I didn't expect you to drop support altogether. Thank you very much for being so patient with me :)

methane:
Thanks

PR #170, opened 2016-02-12
Excepetions hierarchy

palaviv:
I noticed that when I use the message fallback instead of the cython i get different Exception. On my way to fixing this I tried to improve the exceptions in general.

The branch does the following improvements:

1. All exceptions inherit from MsgpackBaseException.
2. The fallback and the cython throw the same exception type.
3. Added new exception PackOverflowError for backwards  compatibility.

### DIFF
--- a/test/test_limits.py
+++ b/test/test_limits.py
@@ -3,36 +3,42 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 import pytest
 
-from msgpack import packb, unpackb, Packer, Unpacker, ExtType
+from msgpack import packb, unpackb, Packer, Unpacker, ExtType, PackException, PackOverflowError, PackValueError
+from msgpack import UnpackValueError, UnpackException, MsgpackBaseException
 
 
-def test_integer():
+@pytest.mark.parametrize("expected_exception", [OverflowError, ValueError, PackOverflowError,
+                                                PackException, PackValueError, MsgpackBaseException])
+def test_integer(expected_exception):
     x = -(2 ** 63)
     assert unpackb(packb(x)) == x
-    with pytest.raises((OverflowError, ValueError)):
+    with pytest.raises(expected_exception):
         packb(x-1)
 
     x = 2 ** 64 - 1
     assert unpackb(packb(x)) == x
-    with pytest.raises((OverflowError, ValueError)):
+    with pytest.raises(expected_exception):
         packb(x+1)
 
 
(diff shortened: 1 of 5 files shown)


methane:
> All exceptions inherit from MsgpackBaseException.

I don't like it.  Why base exception is required?

```
In [1]: import json

In [2]: json.loads("xyzzy")
JSONDecodeError: Expecting value: line 1 column 1 (char 0)

In [3]: json.JSONDecodeError.__mro__
Out[3]: (json.decoder.JSONDecodeError, ValueError, Exception, BaseException, object)
```

There are no BaseJsonException in json module.

palaviv:
I think it is easier for a user when there is a base exception if he wants to catch all exception that are related to the specific module. I think maybe changing it to MsgpackException from MsgpackBaseException will be better.
For example in requests all exceptions inherit from requests.exceptions.RequestException (http://docs.python-requests.org/en/master/user/quickstart/#errors-and-exceptions).

methane:
> I think it is easier for a user when there is a base exception if he wants to catch all exception that are related to the specific module.

He should catch just `Exception`.

In case of requests module, there are many completely different reasons to fail to request: lookup, connect, HTTP error, timeout, etc...
In case of msgpack, I don't think there is enough reason to choose different API design from json.

methane:
All difference between msgpack module and json module should be coming from difference between msgpack and json.
If you have confidence about adding root exception is better, please add it to json before msgpack.

palaviv:
I thought the root exception will be something nice to add (I don't see any down side) but I can see why msgpack is more like json and less like requests.
I removed the MsgpackBaseException and the tests are working now.

test/test_limits.py
```diff
@@ -3,36 +3,42 @@
...
 
 
-def test_integer():
+@pytest.mark.parametrize("expected_exception", [OverflowError, ValueError, PackOverflowError,
+                                                PackException, PackValueError])
+def test_integer(expected_exception):
     x = -(2 ** 63)
     assert unpackb(packb(x)) == x
-    with pytest.raises((OverflowError, ValueError)):
+    with pytest.raises(expected_exception):
```
methane: this parameterize is too much.
Test only with most concrete exception type.

methane:
> I thought the root exception will be something nice to add (I don't see any down side)

More public names is significant downside.
It makes harder to maintainance and performance tuning.

Issue #126, opened 2015-02-20
memoryview objects are not fully supported


himikof:
`memoryview` is the python3 type for non-owning memory buffer objects, also backported to python 2.7. `unpackb` and `Unpacker.feed` should unpack them without copying, and packing functions should handle them the same as `bytes`.
The current support is quite limited due to multiple issues:
### Python 3.3
```
>>> msgpack._unpacker.unpackb(memoryview(b'\x91\xc3'))
[True]
>>> msgpack.fallback.unpackb(memoryview(b'\x91\xc3'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/nofitserov/.local/lib64/python3.3/site-packages/msgpack/fallback.py", line 93, in unpackb
    ret = unpacker._fb_unpack()
  File "/home/nofitserov/.local/lib64/python3.3/site-packages/msgpack/fallback.py", line 383, in _fb_unpack
    typ, n, obj = self._read_header(execute, write_bytes)
  File "/home/nofitserov/.local/lib64/python3.3/site-packages/msgpack/fallback.py", line 274, in _read_header
    b = ord(c)
TypeError: ord() expected string of length 1, but memoryview found
>>> msgpack._packer.Packer().pack(memoryview(b'abc'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "_packer.pyx", line 224, in msgpack._packer.Packer.pack (msgpack/_packer.cpp:224)
  File "_packer.pyx", line 226, in msgpack._packer.Packer.pack (msgpack/_packer.cpp:226)
  File "_packer.pyx", line 221, in msgpack._packer.Packer._pack (msgpack/_packer.cpp:221)
TypeError: can't serialize <memory at 0x7f37c3e41460>
>>> msgpack.fallback.Packer().pack(memoryview(b'abc'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/nofitserov/.local/lib64/python3.3/site-packages/msgpack/fallback.py", line 618, in pack
    self._pack(obj)
  File "/home/nofitserov/.local/lib64/python3.3/site-packages/msgpack/fallback.py", line 615, in _pack
    raise TypeError("Cannot serialize %r" % obj)
TypeError: Cannot serialize <memory at 0x7f37c3e41390>
```
### Python 2.7
```
>>> msgpack._unpacker.unpackb(memoryview(b'\x91\xc3'))
[...]

PR #182, opened 2016-04-28
Fix problems associated with packing memoryviews

jfolz:
Hi,

This PR addresses issues pointed out in https://github.com/msgpack/msgpack-python/issues/126#issuecomment-214425711

- The pure python fallback used the length, not the number of bytes of a memoryview. Views with multi-byte types would be prefixed with a fraction of their real length.
- Packing without ``use_bin_type=True`` and unpacking with encoding set leads to errors. ``use_bin_type=True`` is now enforced for memoryviews to avoid this.
- Add some tests that pack/unpack memoryviews of different types and sizes.

### DIFF
--- a/msgpack/_packer.pyx
+++ b/msgpack/_packer.pyx
@@ -39,6 +39,7 @@ cdef extern from "pack.h":
     int msgpack_pack_ext(msgpack_packer* pk, char typecode, size_t l)
 
 cdef int DEFAULT_RECURSE_LIMIT=511
+cdef size_t ITEM_LIMIT = (2**32)-1
 
 
 cdef class Packer(object):
@@ -178,7 +179,7 @@ cdef class Packer(object):
                    ret = msgpack_pack_double(&self.pk, dval)
             elif PyBytes_CheckExact(o) if strict_types else PyBytes_Check(o):
                 L = len(o)
-                if L > (2**32)-1:
+                if L > ITEM_LIMIT:
                     raise PackValueError("bytes is too large")
                 rawval = o
                 ret = msgpack_pack_bin(&self.pk, L)
@@ -189,7 +190,7 @@ cdef class Packer(object):
                     raise TypeError("Can't encode unicode string: no encoding is specified")
                 o = PyUnicode_AsEncodedString(o, self.encoding, self.unicode_errors)
                 L = len(o)
-                if L > (2**32)-1:
+                if L > ITEM_LIMIT:
                     raise PackValueError("unicode string is too large")
                 rawval = o
                 ret = msgpack_pack_raw(&self.pk, L)
@@ -198,7 +199,7 @@ cdef class Packer(object):
             elif PyDict_CheckExact(o):
                 d = <dict>o
                 L = len(d)
-                if L > (2**32)-1:
+                if L > ITEM_LIMIT:
                     raise PackValueError("dict is too large")
                 ret = msgpack_pack_map(&self.pk, L)
                 if ret == 0:
@@ -209,7 +210,7 @@ cdef class Packer(object):
                         if ret != 0: break
             elif not strict_types and PyDict_Check(o):
                 L = len(o)
-                if L > (2**32)-1:
+                if L > ITEM_LIMIT:
                     raise PackValueError("dict is too large")
                 ret = msgpack_pack_map(&self.pk, L)
                 if ret == 0:
@@ -223,13 +224,13 @@ cdef class Packer(object):
                 longval = o.code
                 rawval = o.data
                 L = len(o.data)
-                if L > (2**32)-1:
+                if L > ITEM_LIMIT:
                     raise PackValueError("EXT data is too large")
                 ret = msgpack_pack_ext(&self.pk, longval, L)
                 ret = msgpack_pack_raw_body(&self.pk, rawval, L)
             elif PyList_CheckExact(o) if strict_types else (PyTuple_Check(o) or PyList_Check(o)):
                 L = len(o)
-                if L > (2**32)-1:
+                if L > ITEM_LIMIT:
                     raise PackValueError("list is too large")
                 ret = msgpack_pack_array(&self.pk, L)
                 if ret == 0:
@@ -240,7 +241,7 @@ cdef class Packer(object):
                 if PyObject_GetBuffer(o, &view, PyBUF_SIMPLE) != 0:
                     raise PackValueError("could not get buffer for memoryview")
                 L = view.len
-                if L > (2**32)-1:
+                if L > ITEM_LIMIT:
                     PyBuffer_Release(&view);
                     raise PackValueError("memoryview is too large")
                 ret = msgpack_pack_bin(&self.pk, L)
@@ -271,8 +272,8 @@ cdef class Packer(object):
         msgpack_pack_ext(&self.pk, typecode, len(data))
         msgpack_pack_raw_body(&self.pk, data, len(data))
 
-    def pack_array_header(self, long long size):
-        if size > (2**32-1):
+    def pack_array_header(self, size_t size):
+        if size > ITEM_LIMIT:
             raise PackValueError
         cdef int ret = msgpack_pack_array(&self.pk, size)
         if ret == -1:
@@ -284,8 +285,8 @@ cdef class Packer(object):
             self.pk.length = 0
             return buf
 
-    def pack_map_header(self, long long size):
-        if size > (2**32-1):
+    def pack_map_header(self, size_t size):
+        if size > ITEM_LIMIT:
             raise PackValueError
         cdef int ret = msgpack_pack_map(&self.pk, size)
         if ret == -1:
--- a/msgpack/fallback.py
+++ b/msgpack/fallback.py
@@ -685,7 +685,7 @@ class Packer(object):
                     default_used = True
                     continue
                 raise PackOverflowError("Integer value out of range")
-            if self._use_bin_type and check(obj, (bytes, memoryview)):
+            if self._use_bin_type and check(obj, bytes):
                 n = len(obj)
                 if n <= 0xff:
                     self._buffer.write(struct.pack('>BB', 0xc4, n))
@@ -696,7 +696,7 @@ class Packer(object):
                 else:
                     raise PackValueError("Bytes is too large")
                 return self._buffer.write(obj)
-            if check(obj, (Unicode, bytes, memoryview)):
+            if check(obj, (Unicode, bytes)):
                 if check(obj, Unicode):
                     if self._encoding is None:
                         raise TypeError(
@@ -715,6 +715,28 @@ class Packer(object):
                 else:
                     raise PackValueError("String is too large")
                 return self._buffer.write(obj)
+            if check(obj, memoryview):
+                n = len(obj) * obj.itemsize
+                if self._use_bin_type:
+                    if n <= 0xff:
+                        self._buffer.write(struct.pack('>BB', 0xc4, n))
+                    elif n <= 0xffff:
+                        self._buffer.write(struct.pack(">BH", 0xc5, n))
+                    elif n <= 0xffffffff:
+                        self._buffer.write(struct.pack(">BI", 0xc6, n))
+                    else:
+                        raise PackValueError("memoryview is too large")
+                    return self._buffer.write(obj)
+                else:
+                    if n <= 0x1f:
+                        self._buffer.write(struct.pack('B', 0xa0 + n))
+                    elif n <= 0xffff:
+                        self._buffer.write(struct.pack(">BH", 0xda, n))
+                    elif n <= 0xffffffff:
+                        self._buffer.write(struct.pack(">BI", 0xdb, n))
+                    else:
+                        raise PackValueError("memoryview is too large")
+                    return self._buffer.write(obj)
             if check(obj, float):
                 if self._use_float:
                     return self._buffer.write(struct.pack(">Bf", 0xca, obj))
--- a/msgpack/fallback.py
+++ b/msgpack/fallback.py
@@ -685,58 +685,29 @@ class Packer(object):
                     default_used = True
                     continue
                 raise PackOverflowError("Integer value out of range")
-            if self._use_bin_type and check(obj, bytes):
+            if check(obj, bytes):
                 n = len(obj)
-                if n <= 0xff:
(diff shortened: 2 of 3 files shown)


jfolz:
Regarding the ITEM_LIMIT:
Looking at the Cython output I saw the ``2**32-1`` was causing unnecessary Python API calls. Couldn't resist making it constant.
Sorry :)

msgpack/fallback.py
```diff
@@ -715,6 +715,20 @@ def _pack(self, obj, nest_limit=DEFAULT_RECURSE_LIMIT,
...
+                        self._buffer.write(struct.pack('>BB', 0xc4, n))
+                    elif n <= 0xffff:
+                        self._buffer.write(struct.pack(">BH", 0xc5, n))
+                    elif n <= 0xffffffff:
+                        self._buffer.write(struct.pack(">BI", 0xc6, n))
+                    else:
+                        raise ValueError("memoryview is too large")
+                    return self._buffer.write(obj)
+                else:
+                    raise ValueError("need use_bin_type=True to pack memoryview")
```
ThomasWaldmann: can you say why precisely? is this a bug?

test/test_memoryview.py
```diff
@@ -2,11 +2,79 @@
...
+
+def test_pack_memoryview_float_65535():
+    data = array('f', range(0xffff // 4))
+    view = __make_view(data)
+    reconstructed = __make_array('f', unpackb(packb(view, use_bin_type=True)))
+    assert data == reconstructed
+
+
+def test_pack_memoryview_float_4294967295():
+    data = array('f', range(0x10000 // 4))
```
ThomasWaldmann: func name and range disagree here.

test/test_memoryview.py
```diff
@@ -2,11 +2,79 @@
...
+
+def test_pack_memoryview_byte_65535():
+    data = array('l', range(0xffff // 4))
+    view = __make_view(data)
+    reconstructed = __make_array('l', unpackb(packb(view, use_bin_type=True)))
+    assert data == reconstructed
+
+
+def test_pack_memoryview_byte_4294967295():
+    data = array('l', range(0x10000 // 4))
```
ThomasWaldmann: func name vs. range

ThomasWaldmann:
Side note: I recently tried to make the msgpack calls in borgbackup round-tripping bytes and str (== bytes and unicode, it is python3) cleanly. It worked but in the end I did not go for it because of compatibility with old backup repositories. https://github.com/borgbackup/borg/pull/973

msgpack/fallback.py
jfolz: If encoding is set, unpack will attempt to decode all raw bytes. Since memoryviews are always binary that's an error. It's not necessary to enforce ``use_bin_type=True`` here, but I think it makes a lot of sense.

test/test_memoryview.py
jfolz: Names correspond to number of bytes in the object to trigger different packing branches. I decided against creating a 4GB array for the last one :)

msgpack/fallback.py
ThomasWaldmann: Still not getting it.

If you have bytes object and the content happens to be validly encoded bytes, it will happily decode them. If not, then not.

If you have a memoryview of such a bytes object and the contents happens to be validly encoded bytes, it will also happily decode them. If not, then not.

So, what's the fundamentaly difference?

jfolz: IMO memoryview objects are considered arbitrary binary data and therefore cannot be decoded to Unicode.

jfolz: Maybe it would be better to just tell people to use ``use_bin_type=True`` if they want to unpack with encoding.

ThomasWaldmann: the problem is that bin type was not there ever and not all code is using use_bin_type=True.

so if one wants to create compatible output with "before bin type" code, that must be possible.

such code does also not use the automatic decoding offered by the Unpacker, because b"foo" and u"foo" data is packed into same output bytes. so, as the unpacker can not know what to decode and what to leave as is, the application needs to know that and do it (or not).

msgpack/fallback.py
```diff
@@ -696,7 +696,7 @@ def _pack(self, obj, nest_limit=DEFAULT_RECURSE_LIMIT,
                 else:
                     raise PackValueError("Bytes is too large")
                 return self._buffer.write(obj)
-            if check(obj, (Unicode, bytes, memoryview)):
+            if check(obj, (Unicode, bytes)):
```
ThomasWaldmann: the easiest way to fix your code, is to keep memoryview in that list.
and move the block you added below to above this block, using
```
if self._use_bin_type and check(obj, memoryview):
```

test/test_memoryview.py
```diff
@@ -2,11 +2,79 @@
...
+
+def test_pack_memoryview_no_bin_type():
+    data = array('l', range(0xff))
+    view = __make_view(data)
+    error = None
+    try:
+        packb(view, use_bin_type=False)
+    except ValueError as e:
+        error = e
+    assert isinstance(error, ValueError)
```
ThomasWaldmann: this should be replaced by tests for not-raising packing / unpacking with use_bin_type = False.

msgpack/_packer.pyx
```diff
@@ -224,27 +225,29 @@ cdef class Packer(object):
...
                 if ret == 0:
                     for v in o:
                         ret = self._pack(v, nest_limit-1)
                         if ret != 0: break
             elif PyMemoryView_Check(o):
+                if not self.pk.use_bin_type:
+                    raise ValueError("need use_bin_type=True to pack memoryview")
                 if PyObject_GetBuffer(o, &view, PyBUF_SIMPLE) != 0:
-                    raise PackValueError("could not get buffer for memoryview")
+                    raise ValueError("could not get buffer for memoryview")
```
ThomasWaldmann: hmm, if i understood correctly, PackValueError is only deprecated.

That means, users of msgpack must now change their code to catch ValueError instead of PackValueError. But as long as they still might catch PackValueError, msgpack must not be changed to raise ValueError because PackValueError is more special than ValueError and thus "except PackValueError" won't catch ValueError.

msgpack/_packer.pyx
```diff
@@ -224,27 +225,29 @@ cdef class Packer(object):
...
                 L = view.len
-                if L > (2**32)-1:
+                if L > ITEM_LIMIT:
                     PyBuffer_Release(&view);
-                    raise PackValueError("memoryview is too large")
-                ret = msgpack_pack_bin(&self.pk, L)
+                    raise ValueError("memoryview is too large")
```
ThomasWaldmann: keep PackValueError

msgpack/fallback.py
```diff
@@ -715,6 +715,20 @@ def _pack(self, obj, nest_limit=DEFAULT_RECURSE_LIMIT,
...
+                n = len(obj) * obj.itemsize
+                if self._use_bin_type:
+                    if n <= 0xff:
+                        self._buffer.write(struct.pack('>BB', 0xc4, n))
+                    elif n <= 0xffff:
+                        self._buffer.write(struct.pack(">BH", 0xc5, n))
+                    elif n <= 0xffffffff:
+                        self._buffer.write(struct.pack(">BI", 0xc6, n))
+                    else:
+                        raise ValueError("memoryview is too large")
```
ThomasWaldmann: use PackValueError

msgpack/_packer.pyx
jfolz: You're right. PackValueError should be phased out some time later.
I found other places in Packer that raise plain ValueErrors though, not sure what to do with them.

msgpack/fallback.py
jfolz: I would like to keep memoryviews separate to raise errors with less confusing messages. E.g., I can then replicate the same errors raised in C and Python.
I can also eliminate some unnecessary branches that way.

msgpack/_packer.pyx
jfolz: Also, C code raises PackValueError, oops ;)

jfolz:
@ThomasWaldmann Thanks for reviewing! I think the last commits address your comments.

jfolz:
If anyone is wondering why str8 ist not used without bin type:
https://github.com/msgpack/msgpack/blob/0b8f5ac67cdd130f4d4d4fe6afb839b989fdb86a/spec.md#upgrading-messagepack-specification

I was. Now I know ;)

test/test_memoryview.py
```diff
@@ -2,11 +2,110 @@
...
+        a = array(f)
+        a.frombytes(data)
+        return a
+
+
+def test_pack_memoryview_byte_8():
+    data = array('l', range(0xff // 4))
+    view = __make_view(data)
+    reconstructed = __make_array('l', unpackb(packb(view)))
+    assert data == reconstructed
```
methane: I don't like "reconstructed" approach for this test.
This test will success even if `packb(v, use_bin_type=False)` produce bin.

msgpack/_packer.pyx
```diff
@@ -241,10 +242,10 @@ cdef class Packer(object):
                 if PyObject_GetBuffer(o, &view, PyBUF_SIMPLE) != 0:
                     raise PackValueError("could not get buffer for memoryview")
                 L = view.len
-                if L > (2**32)-1:
+                if L > ITEM_LIMIT:
                     PyBuffer_Release(&view);
                     raise PackValueError("memoryview is too large")
-                ret = msgpack_pack_bin(&self.pk, L)
+                ret = msgpack_pack_raw(&self.pk, L)
```
methane: What this change makes?

jfolz: I misread the code of ``msgpack_pack_raw``, this shouldn't have changed. Reverted it now. Thanks!

test/test_memoryview.py
jfolz: As has been proven when I broke bin type packing with that other change. Can you suggest a better method?

jfolz: I will do it similar these tests:
https://github.com/msgpack/msgpack-python/blob/6b113a6fb37ffb969e92429b06aab9ea9b8eeb4a/test/test_newspec.py
Hopefully I will have some time on Monday.

methane:
I'm sorry, I made conflict against this PR.  Could you merge and fix it?

jfolz:
@methane No problem, I rebased and squashed.

ThomasWaldmann:
I noticed some potential inconsistencies in the code: is the return value of _pack somehow specified?

At some places in the code it looks like (simplified):
```
return write(header + data)
```
Other places:
```
write(header)
return write(data)
```
Other places:
```
return
```

jfolz:
@ThomasWaldmann The Cython part is pretty clear about this: anything other than 0 means an error occurred.
In fallback, sometimes nothing is returned, like ``pack_ext_type``. What is returned if anything will depend on the buffer implementation. It may not be pretty, but since exceptions are raised where they happen and the return type never used/stored, this should not be an issue.

test/test_memoryview.py
```diff
@@ -2,11 +2,134 @@
...
-def test_pack_memoryview():
-    data = bytearray(range(256))
-    view = memoryview(data)
-    unpacked = unpackb(packb(view))
-    assert data == unpacked
+# For Python < 3:
+#  - array type only supports old buffer interface
+#  - array.frombytes is not available, must use deprecated array.fromstring
+if sys.version_info[0] < 3:
+    def __memoryview(obj):
```
methane: Why do you use double underscore prefix here?
Single underscore is enough (while it's not necessary for test packages since there are no public (stable) API in tests).

test/test_memoryview.py
```diff
@@ -2,11 +2,134 @@
...
+    __run_test('B', 31, b'\xbf', b'', False)
+
+
+def test_memoryview_float_fixstr():
+    __run_test('f', 28, b'\xbc', b'', False)
+
+
+# ----------
+# test str16
+# ----------
```
methane: Decoration comment like this is very rarely used in Python community.
I dislike too.

test/test_memoryview.py
```diff
@@ -2,11 +2,134 @@
...
+
+
+def test_memoryview_byte_str16():
```
methane: module name contains "memoryview" already.
`test_str16_from_byte`, `test_str16_from_float` is more readable.

msgpack/fallback.py
methane: Having separated memoryview block is OK.
But such a large duplicated code block seems bad.
Please split code like `msgpack_pack_raw` and `msgpack_pack_bin` in C version.
Exception message can be build by `"%s object is too large" % (type(obj),)`

test/test_memoryview.py
jfolz: Just out of habit. If something is not intended to be used elsewhere, I make it private. I made a ``make_memoryview`` function instead now.
However, I'm not really satisfied with this solution myself. Old-style buffers don't have type information, so this does not test float packing in 2.7 at all. So far if have only been able to get properly typed memoryviews from numpy.

test/test_memoryview.py
jfolz: OK.

test/test_memoryview.py
methane: see also: http://stackoverflow.com/questions/1301346/the-meaning-of-a-single-and-a-double-underscore-before-an-object-name-in-python

ThomasWaldmann:
\o/

Now, after some testing, we just need all the memoryview supporting stuff in a new release. ;)

jfolz:
@ThomasWaldmann Packer is looking good. Maybe ext type could allow memoryview as data. 

Unpacker could use some memoryview love ;) Should be easy and very efficient to just slice a memoryview buffer.

ThomasWaldmann:
@jfolz I need it for borgbackup, so I personally don't need ext types. Would be great if memoryviews were generally supported, so we don't need to copy gigabytes / terabytes around in memory without real need.

jfolz:
@ThomasWaldmann I'm thinking in the direction of packing ndarrays with type etc. defined by ext types :)

methane:
@ThomasWaldmann Generally speaking, msgpack is not suitable for serializing gigabytes / terabytes data.
msgpack is like JSON or JSON Lines.
msgpack-python should be optimized for large number of small~medium data chunks.

ThomasWaldmann:
It's not in 1 piece. Our chunks are usually ~2MB.

PR #194, opened 2016-05-09
[wip] msgpack 0.5 experimental implementations

methane:


msgpack/_packer.pyx
```diff
@@ -93,12 +101,17 @@ cdef class Packer(object):
         self.pk.length = 0
 
     def __init__(self, default=None, encoding='utf-8', unicode_errors='strict',
-                 use_single_float=False, bint autoreset=1, bint use_bin_type=0,
+                 use_single_float=False, bint autoreset=1, use_bin_type=None,
                  bint strict_types=0):
+        cdef bint bintype = use_bin_type
+        if use_bin_type is None:
+            PyErr_WarnEx(Warning,
```
jfolz: Python 2 documentation states that, depending on user settings, ``PyErr_WarnEx`` can raise an exception instead of a warning:

https://docs.python.org/2/c-api/exceptions.html#c.PyErr_WarnEx

In that case we should ``return NULL`` here.

For some reason that paragraph is gone in the Python 3 version of the document:

https://docs.python.org/3/c-api/exceptions.html#c.PyErr_WarnEx

jfolz: Apparently Cython does this for you, which is nice.

PR #195, opened 2016-05-12
Use new buffer interface to unpack

jfolz:
This PR adds support for unpacking/feeding from any object that supports the new buffer interface.

For compatibility, an attempt is made to use the old buffer interface if that fails. On success, a ``RuntimeWarning`` is issued to inform users of possible errors and future removal of this feature.

### DIFF
--- a/msgpack/_unpacker.pyx
+++ b/msgpack/_unpacker.pyx
@@ -8,16 +8,23 @@ from cpython.bytes cimport (
 )
 from cpython.buffer cimport (
     Py_buffer,
-    PyBuffer_Release,
+    PyObject_CheckBuffer,
     PyObject_GetBuffer,
+    PyBuffer_Release,
+    PyBuffer_IsContiguous,
+    PyBUF_READ,
     PyBUF_SIMPLE,
+    PyBUF_FULL_RO,
 )
 from cpython.mem cimport PyMem_Malloc, PyMem_Free
 from cpython.object cimport PyCallable_Check
+from cpython.ref cimport Py_DECREF
+from cpython.exc cimport PyErr_WarnEx
 
 cdef extern from "Python.h":
     ctypedef struct PyObject
     cdef int PyObject_AsReadBuffer(object o, const void** buff, Py_ssize_t* buf_len) except -1
+    object PyMemoryView_GetContiguous(object obj, int buffertype, char order)
 
 from libc.stdlib cimport *
 from libc.string cimport *
@@ -110,6 +117,42 @@ cdef inline init_ctx(unpack_context *ctx,
 def default_read_extended_type(typecode, data):
     raise NotImplementedError("Cannot decode extended type with typecode=%d" % typecode)
 
+cdef inline int get_data_from_buffer(object obj,
+                                     Py_buffer *view,
+                                     char **buf,
+                                     Py_ssize_t *buffer_len,
+                                     int *new_protocol) except 0:
+    cdef object contiguous
+    cdef Py_buffer tmp
+    if PyObject_CheckBuffer(obj):
+        new_protocol[0] = 1
+        if PyObject_GetBuffer(obj, view, PyBUF_FULL_RO) == -1:
+            raise
+        if view.itemsize != 1:
+            PyBuffer_Release(view)
+            raise BufferError("cannot unpack from multi-byte object")
+        if PyBuffer_IsContiguous(view, 'A') == 0:
+            PyBuffer_Release(view)
+            # create a contiguous copy and get buffer
+            contiguous = PyMemoryView_GetContiguous(obj, PyBUF_READ, 'C')
+            PyObject_GetBuffer(contiguous, view, PyBUF_SIMPLE)
+            # view must hold the only reference to contiguous,
+            # so memory is freed when view is released
+            Py_DECREF(contiguous)
+        buffer_len[0] = view.len
+        buf[0] = <char*> view.buf
+        return 1
+    else:
+        new_protocol[0] = 0
+        if PyObject_AsReadBuffer(obj, <const void**> buf, buffer_len) == -1:
+            raise BufferError("could not get memoryview")
+        PyErr_WarnEx(RuntimeWarning,
+                     "using old buffer interface to unpack %s; "
+                     "this leads to unpacking errors if slicing is used and "
+                     "will be removed in a future version" % type(obj),
+                     1)
+        return 1
+
 def unpackb(object packed, object object_hook=None, object list_hook=None,
             bint use_list=1, encoding=None, unicode_errors="strict",
             object_pairs_hook=None, ext_hook=ExtType,
@@ -129,27 +172,34 @@ def unpackb(object packed, object object_hook=None, object list_hook=None,
     cdef Py_ssize_t off = 0
     cdef int ret
 
-    cdef char* buf
+    cdef Py_buffer view
+    cdef char* buf = NULL
     cdef Py_ssize_t buf_len
     cdef char* cenc = NULL
     cdef char* cerr = NULL
+    cdef int new_protocol = 0
 
-    PyObject_AsReadBuffer(packed, <const void**>&buf, &buf_len)
+    get_data_from_buffer(packed, &view, &buf, &buf_len, &new_protocol)
 
-    if encoding is not None:
-        if isinstance(encoding, unicode):
-            encoding = encoding.encode('ascii')
-        cenc = PyBytes_AsString(encoding)
+    try:
+        if encoding is not None:
+            if isinstance(encoding, unicode):
+                encoding = encoding.encode('ascii')
+            cenc = PyBytes_AsString(encoding)
 
-    if unicode_errors is not None:
-        if isinstance(unicode_errors, unicode):
-            unicode_errors = unicode_errors.encode('ascii')
-        cerr = PyBytes_AsString(unicode_errors)
+        if unicode_errors is not None:
+            if isinstance(unicode_errors, unicode):
+                unicode_errors = unicode_errors.encode('ascii')
+            cerr = PyBytes_AsString(unicode_errors)
 
-    init_ctx(&ctx, object_hook, object_pairs_hook, list_hook, ext_hook,
-             use_list, cenc, cerr,
-             max_str_len, max_bin_len, max_array_len, max_map_len, max_ext_len)
-    ret = unpack_construct(&ctx, buf, buf_len, &off)
+        init_ctx(&ctx, object_hook, object_pairs_hook, list_hook, ext_hook,
+                 use_list, cenc, cerr,
+                 max_str_len, max_bin_len, max_array_len, max_map_len, max_ext_len)
+        ret = unpack_construct(&ctx, buf, buf_len, &off)
+    finally:
+        if new_protocol:
+            PyBuffer_Release(&view);
+
     if ret == 1:
         obj = unpack_data(&ctx)
         if off < buf_len:
@@ -335,14 +385,20 @@ cdef class Unpacker(object):
     def feed(self, object next_bytes):
         """Append `next_bytes` to internal buffer."""
         cdef Py_buffer pybuff
+        cdef int new_protocol = 0
+        cdef char* buf
+        cdef Py_ssize_t buf_len
+
         if self.file_like is not None:
             raise AssertionError(
                     "unpacker.feed() is not be able to use with `file_like`.")
-        PyObject_GetBuffer(next_bytes, &pybuff, PyBUF_SIMPLE)
+
+        get_data_from_buffer(next_bytes, &pybuff, &buf, &buf_len, &new_protocol)
         try:
-            self.append_buffer(<char*>pybuff.buf, pybuff.len)
+            self.append_buffer(buf, buf_len)
         finally:
-            PyBuffer_Release(&pybuff)
+            if new_protocol:
+                PyBuffer_Release(&pybuff)
 
     cdef append_buffer(self, void* _buf, Py_ssize_t _buf_len):
         cdef:
--- a/msgpack/fallback.py
+++ b/msgpack/fallback.py
@@ -1,8 +1,8 @@
 """Fallback pure Python implementation of msgpack"""
 
 import sys
-import array
 import struct
+import warnings
 
 if sys.version_info[0] == 3:
     PY3 = True
@@ -46,6 +46,7 @@ else:
     from io import BytesIO as StringIO
     newlist_hint = lambda size: []
 
+
 from msgpack.exceptions import (
     BufferFull,
     OutOfData,
@@ -77,6 +78,24 @@ def _check_type_strict(obj, t, type=type, tuple=tuple):
         return type(obj) in t
     else:
         return type(obj) is t
+
+
+def _get_data_from_buffer(obj):
+    try:
+        view = memoryview(obj)
+    except TypeError:
+        # try to use legacy buffer protocol if 2.7, otherwise re-raise
+        if not PY3:
+            view = memoryview(buffer(obj))
+            warnings.warn("using old buffer interface to unpack %s; "
+                          "this leads to unpacking errors if slicing is used and "
+                          "will be removed in a future version" % type(obj),
+                          RuntimeWarning)
+        else:
+            raise
+    if view.itemsize != 1:
+        raise ValueError("cannot unpack from multi-byte object")
+    return view
 
 
 def unpack(stream, **kwargs):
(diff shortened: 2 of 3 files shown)


jfolz:
As for the "finish" part, I can think of one more thing: allow ``ExtType.data`` to be ``memoryview``.

But then I've _really_ run out of things to do with memoryviews for the time being.

methane:
> Default to the new buffer interface methods to access data in Cython unpackb

Why?
I think old buffer API can be used for objects supporting new buffer protocol.  Am I wrong?
Is there sample code shows some object which current Cython unpackb can't accept?

jfolz:
For 3.x, that is correct, as described here: https://docs.python.org/3/c-api/objbuffer.html

As always 2.7 is the culprit. Since both protocols exist there, objects have to support both separately. Ironically, ``memoryview`` is one of the objects that does only supports the new protocol. It's simple to test:
```
>>> buffer(memoryview(b'abc'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: buffer object expected
```
Besides technical reasons, since these functions are deprecated since 3.0, we should IMO refrain from using them if possible. The

methane:
@jfolz Thanks.  How do you think dropping old buffer protocol?
I want to keep code simple as possible, especially Cython codes.

methane:
* msgpack 0.5 --- Add memoryview support and deprecate old buffer protocol (warn if possible).
* msgpack 1.0 --- Remove old buffer protocol support.

jfolz:
@methane Simple code is good code :)

To sum it up:
- Only ``unpackb`` uses it the old protocol
- The only (semi-)relevant builtin type that doesn't support the new protocol is ``array.array``

Very few people will ever see a warning, but backward compatibility is kept. Even less people will have to change their custom types that export the old protocol.

msgpack/_unpacker.pyx
```diff
@@ -145,6 +147,9 @@ def unpackb(object packed, object object_hook=None, object list_hook=None,
         else:
             raise
     else:
+        warnings.warn("unpacking %s requires old buffer protocol,"
```
jfolz: Should this warning be issued after the attempt to get the buffer? If the object does not support any protocol, it will still result in a ``TypeError``, but the message may be a bit confusing.

msgpack/fallback.py
```diff
@@ -259,7 +259,8 @@ def feed(self, next_bytes):
                               DeprecationWarning)
             else:
                 raise
-        assert view.itemsize == 1
+        if view.itemsize != 1:
+            raise ValueError("cannot unpack from multi-byte object")
```
jfolz: I decided to raise a proper exception with explanation here. Multi-byte types are very problematic for fallback code, since there is not way to "cast" ``memoryview`` to a different type.

Usually I'm all for developer responsibility, but honestly, if you're trying to unpack floats, I'm pretty sure you're doing it wrong.

msgpack/_unpacker.pyx
```diff
@@ -141,10 +141,11 @@ def unpackb(object packed, object object_hook=None, object list_hook=None,
     if PyObject_CheckBuffer(packed):
         buffer_supported = 1
         if PyObject_GetBuffer(packed, &view, PyBUF_SIMPLE) == 0:
-            buf = <char*> view.buf
+            if view.itemsize != 1:
+                PyBuffer_Release(&view)
+                raise ValueError("cannot unpack from multi-byte object")
```
jfolz: No surprises, so Cython stays in line with fallback here: multi-byte types not allowed.

jfolz:
These last few commits improve the consistency between Cython and fallback.

Something that's still inconsistent is the behavior of ``Unpack.feed``:

Fallback attempts to use the old buffer protocol if possible/necessary and does not accept multi-byte memoryviews. Cython code only uses the new protocol and accepts multi-byte types. Should Cython be aligned with fallback here?

msgpack/_unpacker.pyx
```diff
@@ -129,12 +131,27 @@ def unpackb(object packed, object object_hook=None, object list_hook=None,
...
+        if PyObject_GetBuffer(packed, &view, PyBUF_SIMPLE) == 0:
+            if view.itemsize != 1:
+                PyBuffer_Release(&view)
+                raise ValueError("cannot unpack from multi-byte object")
+            buf_len = view.len
+            buf = <char*> view.buf
+    else:
+        PyObject_AsReadBuffer(packed, <const void**>&buf, &buf_len)
+        PyErr_WarnEx(DeprecationWarning,
+                     "unpacking %s requires old buffer protocol,"
```
ThomasWaldmann: maybe add a blank at the end of first string.

msgpack/_unpacker.pyx
```diff
@@ -129,12 +131,27 @@ def unpackb(object packed, object object_hook=None, object list_hook=None,
...
+                     "which will be removed in msgpack 1.0" % type(packed),
```
ThomasWaldmann: and a dot at end of sentence.

msgpack/fallback.py
```diff
@@ -246,16 +247,25 @@ def __init__(self, file_like=None, read_size=0, use_list=True,
...
-        if (self._fb_buf_n + len(next_bytes) - self._fb_sloppiness
-                        > self._max_buffer_size):
+        try:
+            view = memoryview(next_bytes)
+        except TypeError:
+            # try to use legacy buffer protocol if 2.7, otherwise re-raise
+            if not PY3:
+                view = memoryview(buffer(next_bytes))
+                warnings.warn("unpacking %s requires old buffer protocol,"
+                              "which will be removed in msgpack 1.0" % type(next_bytes),
```
ThomasWaldmann: same here

msgpack/_unpacker.pyx
methane: Is there any reference about convention of exception message?
```
>>> "3" + 2
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Can't convert 'int' object to str implicitly
```

in `_iomodule.c` file:

```
    /* Parameters validation */
    if (universal) {
        if (creating || writing || appending || updating) {
            PyErr_SetString(PyExc_ValueError,
                            "mode U cannot be combined with x', 'w', 'a', or '+'");
            return NULL;
        }
        if (PyErr_WarnEx(PyExc_DeprecationWarning,
                         "'U' mode is deprecated", 1) < 0)
            return NULL;
        reading = 1;
    }

    if (text && binary) {
        PyErr_SetString(PyExc_ValueError,
                        "can't have text and binary mode at once");
        return NULL;
    }

    if (creating + reading + writing + appending > 1) {
        PyErr_SetString(PyExc_ValueError,
                        "must have exactly one of create/read/write/append mode");
        return NULL;
    }

    if (binary && encoding != NULL) {
        PyErr_SetString(PyExc_ValueError,
                        "binary mode doesn't take an encoding argument");
        return NULL;
    }

    if (binary && errors != NULL) {
        PyErr_SetString(PyExc_ValueError,
                        "binary mode doesn't take an errors argument");
```

So I feel last dot is omitted by convention, at least one line message.

jfolz: I don't know if there is any convention, but if the message is a whole sentence it does feel right to add proper punctuation.

methane: Error messages doesn't start with upper case.  It means it isn't whole sentence.
FWIW, in Golang:

> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context. That is, use fmt.Errorf("something bad") not fmt.Errorf("Something bad"), so that log.Print("Reading %s: %v", filename, err) formats without a spurious capital letter mid-message. This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

I grepped CPython's Lib directry and I feel Python has same convention.

jfolz:
Last commit moves buffer access in Cython to a separate function to simplify code as well as guarantee unpack/unpackb/feed behave identically.

msgpack/_unpacker.pyx
```diff
@@ -20,6 +19,7 @@ from cpython.exc cimport PyErr_WarnEx
 cdef extern from "Python.h":
     ctypedef struct PyObject
     cdef int PyObject_AsReadBuffer(object o, const void** buff, Py_ssize_t* buf_len) except -1
+    cdef int PyObject_GetBuffer(object o, Py_buffer *view, int flags) except -1
```
methane: Why do you move this?

jfolz: Exceptions were ignored (e.g., if array is not C-contiguous), which led to access violations.
Should I move related functions as well?

methane: > Exceptions were ignored

Why?

https://github.com/cython/cython/blob/970c2fc0e676ca22016e14147ada0edba937dc6b/Cython/Includes/cpython/buffer.pxd#L31
Cython's declaration has `except -1`.  I don't know why exception were ignored.

jfolz: You are right. I reverted this change and it had no effect.
At first I could not reproduce the problem, but I think the control flow was wrong, because Cython does not automatically exit a ``cdef`` function when exceptions occur. Checking for errors and explicitly re-raising does the job.

jfolz:
These commits should make memory handling more solid.

The following kinds of data are rejected:

- Multi-byte types
- Multidimensional data
- Sliced data

What I'm not satisfied with:

- A test for contiguity rejection (commented out) does not run on 2.7, because ``memoryview`` slicing with step is not implemented

Overall, a lot of added/moved code. This was a lot less finished than I originally thought, really sorry about that.

methane:
Thanks for your job. I'll review in this week hopefully.

But my first impression is...  I forget why we need such a big code.
Basically, accepting binary data is easy.   Pass `y*` to `PyArg_ParseTuple`.
https://docs.python.org/3/c-api/arg.html

I'm afraid about we're over engineering.

methane:
WHY, This is the code in getargs.c in cpython:

```c

static int
getbuffer(PyObject *arg, Py_buffer *view, const char **errmsg)
{
    if (PyObject_GetBuffer(arg, view, PyBUF_SIMPLE) != 0) {
        *errmsg = "bytes-like object";
        return -1;
    }
    if (!PyBuffer_IsContiguous(view, 'C')) {
        PyBuffer_Release(view);
        *errmsg = "contiguous buffer";
        return -1;
    }
    return 0;
}
```

jfolz:
I agree that the buffer code is more complex than it could be.
However, other than the contiguity and itemsize checks I don't see any opportunities to make it any simpler. Half of the code is need for the legacy protocol to prevent regressions.

On contiguity checks: Sadly there is no simple way to check for it in 2.7/PyPy. I could add a pure Python port of the C code that does these checks, though that would be a lot of code:

https://github.com/python/cpython/blob/master/Objects/abstract.c

methane:
I think there are no reasy about rejecting data which `file.write()` accepts.
Isn't `PyBuffer_IsContiguous()` enough?

jfolz:
Technically it is enough, but my main goal with this was to keep consistency between different runtimes.

Idea: Fallback should be fine with sliced data. For C-code where we need contiguous data, we can try to get it using ``PyBuffer_FromContiguous``.

methane:
One important purpose of fallback module is supporting PyPy.
Have you benchmarked memoryview based unpacker on PyPy?

msgpack/fallback.py
```diff
@@ -246,16 +277,13 @@ def __init__(self, file_like=None, read_size=0, use_list=True,
...
-        if (self._fb_buf_n + len(next_bytes) - self._fb_sloppiness
-                        > self._max_buffer_size):
+        view = _get_data_from_buffer(next_bytes, 1)
+        L = len(view)
+        if self._fb_buf_n + L - self._fb_sloppiness > self._max_buffer_size:
             raise BufferFull
-        self._fb_buf_n += len(next_bytes)
-        self._fb_buffers.append(next_bytes)
+        self._fb_buf_n += L
+        self._fb_buffers.append(view)
```
methane: Do copy here.
Supporting memoryview makes it possible to avoid memory copy on user side, not here.

```py
>>> import io
>>> f = io.BytesIO(b"foobar")
>>> buf = bytearray(4)
>>> n = f.readinto(buf)
>>>  n
4
>>> buf
bytearray(b'foob')
>>> a = [memoryview(buf)[:n]]
>>> n = f.readinto(buf)
>>> n
2
>>> buf
bytearray(b'arob')
>>> a += [memoryview(buf)[:n]]
>>> a
[<memory at 0x10ee60288>, <memory at 0x10ee60348>]
>>> bytes(a[0])
b'arob'   # Overwritten!!!
```

methane: More realistic sample code of `.feed()` usage with memoryview.

```py
unpacker = msgpack.Unpacker()
buf = bytearray(16*1024)
while True:
    n = f.readinto(buf)
    if n is None:
        break
    unpacker.feed(memoryview(buf)[:n])   # more efficient than buf[:n]
    for obj in unpacker:
        callback(obj)
```

jfolz: Since we will be copying anyway it would make sense to use a single, larger buffer. I experimented in that direction, but it was quite complex.

jfolz:
Good point. We should suspend this PR until I can do that.

jfolz:
Using PyPy3 2.4.0.

PyPi version (v0.4.7)
```
unpacking integers (fallback)    0.132916, 0.069149, 0.068334, 0.067976
unpacking bytes (fallback)       0.102578, 0.073634, 0.073475, 0.073621
unpacking lists (fallback)       0.307713, 0.297894, 0.295595, 0.296820
unpacking dicts (fallback)       0.280711, 0.263227, 0.274489, 0.262999
```

This PR (+copy in feed)
```
unpacking integers (fallback)    0.162345, 0.093845, 0.092142, 0.090567
unpacking bytes (fallback)       0.118643, 0.088598, 0.088944, 0.105694
unpacking lists (fallback)       0.342774, 0.324450, 0.345009, 0.324881
unpacking dicts (fallback)       0.351656, 0.316005, 0.326773, 0.315644
```

This PR (revert to bytes as buffers)
```
unpacking integers (fallback)    0.128763, 0.063703, 0.061099, 0.061204
unpacking bytes (fallback)       0.104795, 0.076680, 0.076643, 0.076292
unpacking lists (fallback)       0.313370, 0.303259, 0.302231, 0.300040
unpacking dicts (fallback)       0.287460, 0.266801, 0.277002, 0.266382
```

Seems like PyPy is well-optimized for bytes, not so much for memoryviews.

methane:
@jfolz See #196.  I haven't benchmarked it yet.
But I think we should rewrite buffer implementation before implement accepting memoryview.

msgpack/_unpacker.pyx
```diff
@@ -110,6 +118,40 @@ cdef inline init_ctx(unpack_context *ctx,
...
+            raise
+        if view.itemsize != 1:
+            PyBuffer_Release(view)
+            raise BufferError("cannot unpack from multi-byte object")
+        if PyBuffer_IsContiguous(view, 'A') == 0:
+            contiguous = PyMemoryView_GetContiguous(obj, PyBUF_READ, 'C')
+            PyObject_GetBuffer(contiguous, &tmp, PyBUF_SIMPLE)
+            PyBuffer_Release(view)
+            PyBuffer_FillInfo(view, contiguous, tmp.buf, tmp.len, 1, 0)
+            Py_DECREF(contiguous)
```
jfolz: What this does:
- ``PyMemoryView_GetContiguous`` creates a  ``memoryview`` object to a contiguous copy of the non-contiguous memory
- Get a ``Py_buffer`` for this contiguous memory and transfer the information to the target buffer
- Release resources as necessary

methane:
Maybe related issue: http://bugs.python.org/issue20699

jfolz:
2.7 is a mess with regards to buffer handling. Working with this stuff makes me want to move to 3.5.

jfolz:
I briefly tested Cython's typed memoryviews. Sadly, they require write access, so they don't work for most bytes-like objects.

jfolz:
@methane Any updates on this?

msgpack/_unpacker.pyx
```diff
@@ -150,6 +195,8 @@ def unpackb(object packed, object object_hook=None, object list_hook=None,
              use_list, cenc, cerr,
              max_str_len, max_bin_len, max_array_len, max_map_len, max_ext_len)
     ret = unpack_construct(&ctx, buf, buf_len, &off)
+    if new_protocol:
+        PyBuffer_Release(&view);
```
methane: try ~ finally is required, since unpack_construct, encoding.encode() and unicode_errors.encode() may raise Exception.

methane:
LGTM, except one comment.

jfolz:
I added try-finally and squashed.

Thanks for the help, looking forward to 0.5 :)

methane:
Thanks

PR #206, opened 2016-08-30
README notes about `encoding` and `use_bin_type`

loisaidasam:
Helpful information to save others time.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -217,71 +217,48 @@ You can change this by using `use_bin_type=True` option in Packer and `encoding=
     ['spam', u'egg']
 
 It's worth noting that ``encoding`` (for ``packb()`` and ``unpackb()`` alike), ``use_bin_type``, and other
-parameters have nothing to do with encoding of the string representation of the packed string itself,
-but rather with the encoding of the final unpacked item.
-
-For example:
+parameters have nothing to do with encoding of the packed representation of the string itself,
+but rather with the encoding of the final unpacked item. In other words, the packed representation
+of the item is always going to be a string, seemingly encoded using `latin1` encoding:
 
 .. code-block:: pycon
 
-    # Packing:
-    >>> import msgpack
-    >>> beyonce = u"Beyoncé"
-    >>> beyonce  # unicode
-    u'Beyonc\xe9'
-    >>> beyonce.encode('utf-8')  # regular string
-    'Beyonc\xc3\xa9'
-    # Packed result is a regular string
-    >>> msgpack.packb(beyonce.encode('utf-8'))
-    '\xa8Beyonc\xc3\xa9'
-    # Packed result is still a regular string
-    >>> msgpack.packb(beyonce.encode('utf-8'), encoding='utf-8')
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -217,71 +217,48 @@ You can change this by using `use_bin_type=True` option in Packer and `encoding=
     ['spam', u'egg']
 
 It's worth noting that ``encoding`` (for ``packb()`` and ``unpackb()`` alike), ``use_bin_type``, and other
-parameters have nothing to do with encoding of the string representation of the packed string itself,
-but rather with the encoding of the final unpacked item.
-
-For example:
+parameters have nothing to do with encoding of the packed representation of the string itself,
+but rather with the encoding of the final unpacked item. In other words, the packed representation
+of the item is always going to be a string, seemingly encoded using `latin1` encoding:
```
methane: Totally incorrect and misleading.

README.rst
```diff
@@ -217,71 +217,48 @@ You can change this by using `use_bin_type=True` option in Packer and `encoding=
...
     
-    # Still not utf-8 encoded
-    >>> msgpack.packb(beyonce.encode('utf-8'), encoding='utf-8', use_bin_type=True).decode('utf-8')
-    Traceback (most recent call last):
-      File "<stdin>", line 1, in <module>
-      File "/path/to/lib/python2.7/encodings/utf_8.py", line 16, in decode
-        return codecs.utf_8_decode(input, errors, True)
-    UnicodeDecodeError: 'utf8' codec can't decode byte 0xc4 in position 0: invalid continuation byte
+    for input in (u"Beyoncé", u"Beyoncé".encode('utf-8')):
+        print "---"
```
methane: Don't use Python 2 in README.  It's legacy language.

README.rst
```diff
@@ -217,71 +217,48 @@ You can change this by using `use_bin_type=True` option in Packer and `encoding=
...
+        print "Input: %s (type=%s)" % (input, type(input))
+        for encoding in ('utf-8', 'latin1'):
+            for params in [{}, {'encoding': 'utf-8'}, {'use_bin_type': True}, {'encoding': 'utf-8', 'use_bin_type': True}]:
```
methane: Too much for README.
README is not full document.  It's a quick view.

README.rst
methane: msgpack is "binary" format.  It's not "latin1 encoded string".  Just a binary.  Not a string.

loisaidasam:
In python 2 (which is not a legacy language), `bytes('foo') == str('foo')`.

In my environment, I guess `latin1` is the default encoding, so the bytes representation (which is a string in python 2) seems to be decodable to unicode using the `latin1` encoding. If you need a unicode representation of the packed string (of bytes), you can get it using that encoding.

That is all.

methane:
>  If you need a unicode representation of the packed string (of bytes), 

It's case by case.
There are many string representation of arbitrary bytes: base64, base85, hex encoding.
You should choose best option for your situation.  But best is binary data as binary data.

And it's totally out-of-scope of msgpack.

PR #212, opened 2016-11-26
(WIP) Add Unpacker.offset := starting offset of current object

### LINKED ISSUE #211: Track current stream offset?
In some applications, namely indexing of msgpack streams, it would be very useful to tell the total offset (not the buffer offset) of the last object returned by the Unpacker iterator. In code:

```py
for packet in data:
    unpacker = Unpacker(...)
    unpacker.feed(data)
    for object in unpacker:
        print(unpacker.offset)
```

Would print for example 0, 1234, 4567, and so on.

enkore:
Fixes #211 

WIP due to bugs in native code. There are likely other bugs in the Python implementation as well, plus some missing stuff.

### DIFF
--- a/msgpack/_unpacker.pyx
+++ b/msgpack/_unpacker.pyx
@@ -256,7 +256,7 @@ cdef class Unpacker(object):
     """
     cdef unpack_context ctx
     cdef char* buf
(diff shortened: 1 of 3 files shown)


enkore:
=#227

PR #214, opened 2017-01-11
fix typos and other cosmetic issues

ThomasWaldmann:
cosmetic issues:
- reST headlines' underline length needs to match the headline length
  (looks like somebody is / was using a proportional font)
- Cython code lines do not need to be terminated with a semicolon
- always use triple-double-quotes for docstrings

I did the PR against release-0.4 branch, hope this is correct to get it fixed in next release and also get it merged from there into master.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -160,7 +160,7 @@ Changes
 
 
 0.2.4
-=======
+=====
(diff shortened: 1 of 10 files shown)


methane:
thanks

PR #227, opened 2017-04-29
Unpacker: add tell()

jfolz:
There seems to be a demand for tracking offset (e.g., for indexing) while unpacking from stream. Before 1.0, it is suggested to use `write_bytes` for that purpose. Since it was really easy and a lot more efficient, this PR adds a `tell()` method to `Unpacker`.

### DIFF
--- a/msgpack/_unpacker.pyx
+++ b/msgpack/_unpacker.pyx
@@ -29,6 +29,7 @@ cdef extern from "Python.h":
 from libc.stdlib cimport *
 from libc.string cimport *
 from libc.limits cimport *
+ctypedef unsigned long long uint64_t
 
 from msgpack.exceptions import (
     BufferFull,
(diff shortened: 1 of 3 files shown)


msgpack/_unpacker.pyx
```diff
@@ -314,6 +315,7 @@ cdef class Unpacker(object):
     cdef object object_hook, object_pairs_hook, list_hook, ext_hook
     cdef object encoding, unicode_errors
     cdef Py_ssize_t max_buffer_size
+    cdef uint64_t stream_offset
```
jfolz: `Py_ssize_t` may be 32bit, which is a bit small even today. 16 Exabyte should be plenty for some time.

jfolz: Should be good for 30+ years (if you can unpack continuously at 16GB/s ;))

enkore:
Thanks!

PR #229, opened 2017-05-17
Packer accepts bytearray objects

### LINKED ISSUE #224: can't serialize bytearray
If value = bytearray(b'\x01\x87h\x05')

then packb(value, use_bin_type=True)

returns the error: can't serialize bytearray(b'\x01\x87h\x05')

Is this a known problem?

jfolz:
Fixes #224.

### DIFF
--- a/msgpack/_packer.pyx
+++ b/msgpack/_packer.pyx
@@ -10,6 +10,8 @@ from msgpack import ExtType
 cdef extern from "Python.h":
 
     int PyMemoryView_Check(object obj)
+    int PyByteArray_Check(object obj)
+    int PyByteArray_CheckExact(object obj)
 
 
 cdef extern from "pack.h":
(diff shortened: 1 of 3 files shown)


jfolz:
Tiny niggle: It now says "Str is too large" instead of "Bytes is too large" in 2.7 if the object exceeds `ITEM_LIMIT`.

msgpack/_packer.pyx
```diff
@@ -174,10 +184,11 @@ cdef class Packer(object):
                 else:
                    dval = o
                    ret = msgpack_pack_double(&self.pk, dval)
-            elif PyBytes_CheckExact(o) if strict_types else PyBytes_Check(o):
+            elif PyBytesLike_CheckExact(o) if strict_types else PyBytesLike_Check(o):
                 L = len(o)
                 if L > ITEM_LIMIT:
-                    raise PackValueError("bytes is too large")
+                    raise PackValueError("%s is too large"
+                                         % type(o).__name__.capitalize())
```
methane: I don't like `capitalize()`.
How about `"%s length is too large" % type(o).__name__`?

jfolz: No problem. `capitalize` is just there because it said `Bytes` instead of `bytes` ;)

methane:
thanks.

PR #232, opened 2017-07-02
packer: add passthrough for already packed data

enkore:
This patch adds a Packer option to allow already msgpacked data as an object. Such data is copied verbatim into the output buffer (=passthrough).

This helps when the application has some (possibly large or complex) msgpack structure it wants to embed in another msgpack structure. Without this patch it has to unpack it first and then repack it again. That is no good for throughput and requires a lot of copies and CPU.

### DIFF
--- a/msgpack/__init__.py
+++ b/msgpack/__init__.py
@@ -17,6 +17,15 @@ class ExtType(namedtuple('ExtType', 'code data')):
         return super(ExtType, cls).__new__(cls, code, data)
 
 
+class Passthrough(object):
+    """Class to pass raw msgpack data into a packer."""
+    def __init__(self, data):
+        self.data = data
+
+    def __bytes__(self):
+        return self.data
+
+
 import os
 if os.environ.get('MSGPACK_PUREPYTHON'):
     from msgpack.fallback import Packer, unpack, unpackb, Unpacker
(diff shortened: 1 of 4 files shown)


jfolz:
Interesting. Can you see this help with packing complex data structures, e.g., Numpy arrays? I'm packing **a lot** of arrays and right now the receiver needs to specifically turn the array specs + data back into an actual array.

enkore:
This only helps if the data is already in msgpack format. It could conceivably help if you want to pack something like a large blob if you're able to write the msgpack object header in front of it (e.g. allocate these extra bytes beforehand).

I'm mostly targeting the use case where some other part of an application has already packed something, and the overhead of unpacking then repacking shall be avoided.

enkore:
I see that the memoryview code has not been included in a release yet, though.

methane:
I don't want to any APIs unless it's useful for 20%+ users.
If Java and Ruby msgpack library (the author is msgpack spec writer) doesn't have the API,
I very unlikely to add the API.

In this case, common practice in msgpack is putting already packed data as
msgpack byte object.  This practice is used in fluentd.
Since msgpack's bytes object is not "escaped" format, this practice is efficient enough.
And this practice doesn't require any special APIs: It can be used among many
languages including Python, Ruby, Java, PHP, etc...

enkore:
It changes the format, though, so it is not possible to use that optimization with a pre-defined format.

methane:
I don't think it's fairly little because you added code to serializer core loop.

I'm thinking about reducing APIs which arn't used for 99% people, not adding them.

enkore:
Ok.

PR #240, opened 2017-08-20
Add pyxbld files so you can do testing much more easily.

fake-name:
This PR adds `.pyxbld` files for the two cython files, which allows their use via Cython's [`pyximport()`](https://github.com/cython/cython/tree/master/pyximport) interface.

This is principally useful for testing/development, but since the pyxbld files aren't bundled when building a proper release, it's harmless in all other cases.

Principally, this means you can use test-scripts like:

```

import pyximport
pyximport.install()

import os
print("Importing msgpack")
import msgpack
print("Imported")

data = os.urandom(1024 * 1024) * 375
print("Data generated. Packing", len(data))
message_bytes = msgpack.packb(data, use_bin_type=True)
print("Packed", len(message_bytes))

```

Assuming the `msgpack` dir is in the same folder as the above script, the two `*.pyx` files are automatically compiled at import-time, leading to much faster test cycles.

### DIFF
--- a/msgpack/_packer.pyxbld
+++ b/msgpack/_packer.pyxbld
@@ -1,0 +1,41 @@
+
+import os.path
+import sys
+
+def make_ext(modname, pyxfilename):
+	from distutils.extension import Extension
+
+	if sys.byteorder == 'big':
+		macros = [('__BIG_ENDIAN__', '1')]
+	else:
+		macros = [('__LITTLE_ENDIAN__', '1')]
+
+	ret = Extension(name=modname,
+					sources=[
+							pyxfilename,
+						],
+					libraries=[
+						],
+					include_dirs = [
+							'.',
+							'./msgpack/',
+						],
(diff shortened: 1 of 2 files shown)


msgpack/pack.h
```diff
@@ -48,7 +48,11 @@ static inline int msgpack_pack_write(msgpack_packer* pk, const char *data, size_
     if (len + l > bs) {
         bs = (len + l) * 2;
         buf = (char*)PyMem_Realloc(buf, bs);
-        if (!buf) return -1;
+        if (!buf)
+        {
+            PyErr_SetString(PyExc_MemoryError, "Out of memory exception!");
```
jfolz: Maybe something more descriptive like "cannot grow internal buffer" is more appropriate.

methane:
Please not merge two change in one pull request.

methane:
ping?

Issue #245, opened 2017-09-28
Can't serialize tuples with strict_types


lbolla:
When using `strict_types=True`, serializing tuples fails:

    >>> msgpack.dumps((1,2, 3), strict_types=True)
    ....
    TypeError: can't serialize (1, 2, 3)

(Using "tip" version of msgpack.)

lbolla:
I figured a possible workaround, using `object_hook`:

    def default(o):
        if isinstance(o, tuple):
            return {
                '__type__': 'tuple',
                'value': list(o),
                }
        raise TypeError(repr(o))


    def convert(o):
        if o.get('__type__') == 'tuple':
            return tuple(o['value'])
        return o


    data = msgpack.packb(t, strict_types=True, default=default)
    expected = msgpack.unpackb(data, object_hook=convert)

    assert expected == t

methane:
It's not "workaround", but "proper way".
Msgpack is like json.  It doesn't have "tuple" and "list". It has just "array".

See #100 and #98.  There were requests to distinguish tuple and list.

lbolla:
Thanks! I've created a PR to document in unittests this procedure: it wasn't clear from the docs for me.
Also, the PR fixes a bug in formatting the error message.

PR #246, opened 2017-09-29
Add unittests to document serialisation of tuples

lbolla:
Also, fix formatting of error message in case of tuple.
See https://github.com/msgpack/msgpack-python/issues/245

### DIFF
--- a/msgpack/fallback.py
+++ b/msgpack/fallback.py
@@ -795,7 +795,7 @@ class Packer(object):
                 obj = self._default(obj)
                 default_used = 1
                 continue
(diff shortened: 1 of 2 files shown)


jfolz:
You can use `use_list=False` in unpack if you want tuples instead of lists.

lbolla:
@jfolz That's not enough. I want to be able to *distinguish* between tuples and lists: `use_list=False` would just deserialise both tuples and lists as tuples.

lbolla:
By the way, the above CI error seems spurious?

PR #277, opened 2018-02-02
Fix unicode_errors encoding

nvllsvm:
0.5.2 introduced a regression in `Packer`.  Specifying only `unicode_errors` without `encoding` results in `self.unicode_errors` *not* being set to the result of `PyBytes_AsString`.

Previously, `unicode_errors` was either set to `NULL` or to the result of `PyBytes_AsString`. This restores that behavior while also keeping the new `NULL` default behavior when both are unspecified.
Original defaults of `utf-8` and `strict` were restored to keep API compatibility until these deprecated options are finally removed.

See https://github.com/neovim/python-client/issues/299 for downstream issues.

### DIFF
--- a/msgpack/_packer.pyx
+++ b/msgpack/_packer.pyx
@@ -127,27 +127,26 @@ cdef class Packer(object):
             if not PyCallable_Check(default):
                 raise TypeError("default must be a callable.")
         self._default = default
-        if encoding is None:
-            if unicode_errors is None:
-                self.encoding = NULL
-                self.unicode_errors = NULL
-            else:
-                self.encoding = "utf_8"
-                self.unicode_errors = unicode_errors
+        if encoding is None and unicode_errors is None:
+            self.encoding = NULL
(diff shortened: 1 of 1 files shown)


methane:
Oh, I never think about passing `unicode_errors` without `encoding`.
I never use `bytes.decode()` or `str.encode()` with errors but without encoding.

Thank you for reporting and pull request.  I'll look it and release 0.5.3 soon.
But I recommend you to not use `unicode_errors` option.

nvllsvm:
FWIW - Both `bytes.decode()` and `str.encode` default to `utf-8` if encoding is unspecified. The docs for Python 3 even advertise this functionality.
https://docs.python.org/3.6/library/stdtypes.html#bytes.decode
https://docs.python.org/3.6/library/stdtypes.html#str.encode

methane:
Hm,  but I never omit it because it's not Python 2 compatible, and `open()`'s default
encoding is not UTF-8 always.

PR #278, opened 2018-02-04
Undeprecate unicode_errors

methane:


methane:
I do not remember why I undeprecate this...

methane:
I found it.
https://github.com/neovim/pynvim/issues/299#issuecomment-362790688

Issue #281, opened 2018-02-06
Unclear type mapping with both packb/unpackb, especially with mixed type sequences


rwarren:
## Problem:

The current/packing unpacking situation is confusing and complex when it comes to dealing with the different binary and string types that will be packed to, or unpacked from, the MessagePack "_str format family_" and "_bin format family_" data types. It is difficult to determine the correct combination for a satisfactory type mapping in all situations.

In addition, the current `msgpack-python` (now `msgpack`) implementations do not have a solution (in either direction) for dealing with data containers that contain both string and binary data types.

For example (on the unpacking/deserialization side), the following byte sequence defines a MessagePack array that contains two elements: a unicode snowman character in utf-8, and an arbitrary byte sequence of [0x00, 0x01, 0x02]:

```
data = b'\x92\xa3\xe2\x98\x83\xc4\x03\x00\x01\x02'
```
What possible combination of `msgpack.unpackb` kwargs can _properly_ unpack this to a two element list containing a python string and a suitable binary type (like `bytearray` in Python 2, and `bytes` in Python 3)? Conversely, how would you generate such a MessagePack structure from python (aside from direct generation as above)?

## Proposal:

Rather than having a collection of effectivey global switches that can be sent to `packb`/`unpackb` (for example: `raw_as_bytes` and `use_bin_type`), it would be better if there were a method for defining an explicit typemap that would be used at a per-element level, and which defined the type mappings to use for both `packb` (from python to the MessagePack protocol) and `unpackb` (from MessagePack to python).

For example, it would be great if `msgpack.unpackb(data_bytes, typemap='ideal'`) would get the "ideal" behaviour I outline in the tables further below. When using the `typemap` switch, packing/unpacking could then work in a per-element way, rather than having issues with mixed-type sequences like the global switches currently do. Possible `typemap` values could be similar to in the columns defined further below: `('ideal', '0.4', '0.5', 'default')`, or somesuch, where `default` would be the default value, and currently equate to `typemap='0.5'`. It would also be illegal (raising `ValueError`) to specify kwargs like `raw_as_bytes` together with a `typemap` specification). I think this proposal will resolve any potential compatibility issues.

This `typemap` kwarg behaviour should be bidirectional. Specifically, there should also be a similar possibility for `msgpack.packb(data, typemap=`ideal`).

[...]

PR #282, opened 2018-02-07
added test case "test_mixed_bin_str" (see issue #281)

rwarren:
This is to ensure that behaviour referenced in #281 doesn't break in the future.

### DIFF
--- a/test/test_case.py
+++ b/test/test_case.py
@@ -100,18 +100,26 @@ def test_match():
 def test_unicode():
     assert unpackb(packb('foobar'), use_list=1) == b'foobar'
 
-def test_bintype_symmetry():
+def test_mixed_bin_str():
+    # This test explicitly checks behaviour when a MessagePack byte sequence has both
+    # "bin family" types and "str family" types.
     # For background on this test see github issue #281
-    #  - basically: packing with use_bin_type=True and unpacking with raw=False
-    #     should be completely symmetric
-    buf = b'\x92\xa3\xe2\x98\x83\xc4\x03\x00\x01\x02'
+    #  - basically: packing with use_bin_type=True and unpacking with raw=False should be
+    #     completely symmetric
+    expected = [u'\u2603', b'\x00\x01\x02']
+    buf = (b'\x92'                 # array with length 2      (fixarray)
+           b'\xa3\xe2\x98\x83'     # str format with length 3 (fixstr) (UTF8 snowman)
+           b'\xc4\x03\x00\x01\x02' # bin format with length 3 (bin8) (0x000102)
+           )
+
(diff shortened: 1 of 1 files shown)


methane:
Have you checked there are no test about that for now?

rwarren:
> Have you checked there are no test about that for now?

Yes.  I could find nothing that explicitly, or properly, checks dealing with mixed bin/str elements in MessagePack protocol stream.

----

More detail...

A proper test for this requires:

1. testing a raw MessagePack byte sequence with mixed bin/str types in it
2. explicit testing of the symmetry when using `use_bin_type=True` and `raw=False`
3. assertion of identical python types (not just equality testing with `==`)
4. testing properly under both python 2 and python 3

The closest I could find (that do not meet these criteria) are:

* tests in `test_case.py`
  * No tests for mixed bin/str types. They also don't test any `use_bin_type` usage. They also use equality testing with `==` which does not ensure type matches.
* `test_stricttype.test_tuple` seems like, at first glance, that it kind of checks it... but it doesn't meet the criteria
  * it does _partially_ cover the condition under python 3 (the right pack/unpack sequence happens to happen there), but there is no explicit byte stream specified, and it doesn't meet several other other criteria (eg: no type checking, doesn't properly cover python2, etc). In addition, the fact that it happens to do a piece of the test is mostly by accident, since it is not the purpose of the test.
* `test_pack.py` and `test_unpack.py ` also don't have anything that properly tests it

With a lack of an explicit test, I thought one should be added, so I wrote it rather than requested it. Note that I also considered writing a test case with multiple cases (similar to what `test_case.test_match` does), but ended up thinking that a clean single case test for the specific conditions was better.

Also - I renamed the test to better reflect what it does. The original name wasn't great.

jfolz:
Mixed testing should not be necessary. Packing/unpacking does not change internal state of the Packer/Unpacker beyond managing message buffers. As long as each object/token is handled correctly any mix of types will be handled correctly.
That said there are some issues with the current test suite. E.g., this assert [test_pack.py#L15](https://github.com/msgpack/msgpack-python/blob/4b72b6177321ef24b6c7af2354fb980db69e2aec/test/test_pack.py#L15) is not triggered if given `str` and `unicode` in Python 2.7. This type of assert is used in several tests.
Also, a lot of tests use some form of packing-unpacking chain. Regressions in the Packer can be hidden by regressions in the Unpacker. They should probably be rewritten in the style of [test_format.py](https://github.com/msgpack/msgpack-python/blob/4b72b6177321ef24b6c7af2354fb980db69e2aec/test/test_format.py): specifying the ground-truth data-object pair as per msgpack spec thas to be handled correctly.

PR #310, opened 2018-09-28
Alternate fixes for jython and legacy c python.

devendor:
The original batch wasn't working for jython and seemed to be more a lack of implicit type casting in older objects and alternate implementation pythons.  This seems less intrusive, more straightforward and works on jython-2.7.1.

### DIFF
--- a/msgpack/fallback.py
+++ b/msgpack/fallback.py
@@ -213,7 +213,7 @@ class Unpacker(object):
             for o in unpacker:
                 process(o)
     """
-    
+
     def __init__(self, file_like=None, read_size=0, use_list=True, raw=True,
                  object_hook=None, object_pairs_hook=None, list_hook=None,
                  encoding=None, unicode_errors=None, max_buffer_size=0,


msgpack/fallback.py
```diff
@@ -212,7 +205,12 @@ class Unpacker(object):
             for o in unpacker:
                 process(o)
     """
-
+    # Some old pythons don't support `self._unpack_from()` `bytearray`. 
+    _unpack_from = staticmethod(
+        struct.unpack_from if sys.version_info >= (2, 7, 6)
+        else lambda f,b,o=0: struct.unpack_from(f, b[:o+2].tobytes(),o)
+    )
```
methane: Use global function instead of staticmethod.

devendor: Sorry it took so many tries.   I have a head cold so it was like typing three lines of code with my toes this morning.

msgpack/fallback.py
```diff
@@ -212,7 +213,7 @@ class Unpacker(object):
             for o in unpacker:
                 process(o)
     """
-
+
```
methane: Don't this

msgpack/fallback.py
```diff
@@ -302,7 +297,7 @@ def feed(self, next_bytes):
             self._buff_i -= self._buf_checkpoint
             self._buf_checkpoint = 0
 
-        self._buffer += view
+        self._buffer += bytearray(view)
```
methane: Doesn't this increase data copy?

devendor: That is an interesting question.  My assumption is that the inplace add / concatenation operator is doing an implicit typecast for you in the background rather than throwing TypeError in newer python, so an explicit typecast was just replacing a step that had to happen implicitly somewhere anyway.  I don't actually want to read a bunch of code to verify that assumption though.  I figured out that .extend() works reliably with implicit typecasting in old and new which should be about an even trade.

PR #319, opened 2018-11-08
unpacker: Make default size limit smaller

### LINKED ISSUE #295: Change default decoder limits
Current default limits seems too large.  It can cause DoS attack.
Change default value to more safe value.

Current plan is "about 1MiB on amd64 system".

* `max_bin_len`: 1024*1024
* `max_str_len`: 1024*1024
* `max_array_len`: 1024*1024/8  (each pointer has 8 bytes)
* `max_map_len`: 1024*1024/32  (8byte key,hash,value, and some extra space)

methane:
To avoid DoS attack, make default size limit smaller.

Fixes #295

### DIFF
--- a/msgpack/_unpacker.pyx
+++ b/msgpack/_unpacker.pyx
@@ -162,11 +162,11 @@ def unpackb(object packed, object object_hook=None, object list_hook=None,
             bint use_list=True, bint raw=True,
             encoding=None, unicode_errors=None,
             object_pairs_hook=None, ext_hook=ExtType,
-            Py_ssize_t max_str_len=2147483647, # 2**32-1
-            Py_ssize_t max_bin_len=2147483647,
-            Py_ssize_t max_array_len=2147483647,
-            Py_ssize_t max_map_len=2147483647,
-            Py_ssize_t max_ext_len=2147483647):
+            Py_ssize_t max_str_len=1024*1024,
+            Py_ssize_t max_bin_len=1024*1024,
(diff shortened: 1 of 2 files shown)


msgpack/_unpacker.pyx
```diff
@@ -261,16 +261,19 @@ cdef class Unpacker(object):
...
     :param int max_array_len:
-        Limits max length of array. (default: 2**31-1)
+        Limits max length of array. (default: 128*1024)
 
     :param int max_map_len:
-        Limits max length of map. (default: 2**31-1)
+        Limits max length of map. (default: 32*1024)
+
+    :param int max_ext_len:
+        Limits max length of map. (default: 1024*1024)
```
decaz: I suppose it's incorrect description.

msgpack/fallback.py
```diff
@@ -184,17 +184,19 @@ class Unpacker(object):
...
     :param int max_array_len:
-        Limits max length of array. (default: 2**31-1)
+        Limits max length of array. (default: 128*1024)
 
     :param int max_map_len:
-        Limits max length of map. (default: 2**31-1)
+        Limits max length of map. (default: 32*1024)
 
+    :param int max_ext_len:
+        Limits max length of map. (default: 1024*1024)
```
decaz: And this is too (incorrect description).

methane: nice catch

PR #347, opened 2019-02-13
Support packing directly into a writer object

huguesb:
When packing larger objects, the current default behavior
of packing the entire object into a buffer and then writing
it into the provided output stream results in massive slowdowns.

To address this use case, Packer now accepts an optional `writer`
kwarg, which must be an object with a `write()` method. When this
argument is given:
  - the pure Python implementation writes directly to this object,
    completely bypassing StringIO
  - the C implementation only writes to this object when its internal
    buffer fills up, and to flush the remainder of the buffer at the
    end of a pack() call. Large elements may also bypass the buffer
    completely.

Add a few more tests to cover large values

### DIFF
--- a/msgpack/__init__.py
+++ b/msgpack/__init__.py
@@ -33,8 +33,11 @@ def pack(o, stream, **kwargs):
 
     See :class:`Packer` for options.
     """
-    packer = Packer(**kwargs)
-    stream.write(packer.pack(o))
+    if 'writer' in kwargs:
+        packer = Packer(**kwargs)
+        stream.write(packer.pack(o))
+    else:
+        Packer(writer=stream, **kwargs).pack(o)
 
 
 def packb(o, **kwargs):
--- a/msgpack/_packer.pyx
+++ b/msgpack/_packer.pyx
@@ -18,6 +18,11 @@ cdef extern from "pack.h":
         size_t length
         size_t buf_size
         bint use_bin_type
+        PyObject* writer
+
+    void msgpack_packer_init(msgpack_packer* pk, PyObject* writer)
+    void msgpack_packer_free(msgpack_packer* pk)
+    int msgpack_pack_flush(msgpack_packer* pk)
 
     int msgpack_pack_int(msgpack_packer* pk, int d)
     int msgpack_pack_nil(msgpack_packer* pk)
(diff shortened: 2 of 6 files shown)


methane:
Packer is complicated already.  Please don't make it more complicated.
(I regret about `file_like` option in Unpacker too.)

Try new class instead; maybe "StreamingPacker".

huguesb:
@methane what's the likelihood that you would accept a pull request with a new class?

methane:
I can't say anything until read the code.
If I merge it, I need to maintain it for a long time.  So I don't accept code I feel I don't want to maintain it.

Even if I don't accept, you can package your new class as new package on PyPI.

jfolz:
I really recommend benchmarking this before putting in too much effort. It might be possible to save memory and reduce latency, but packing is really really fast already. Easily several GB/s when (byte) strings are involved. Larger elements will definitely have to be written in chunks to see any gains.

huguesb:
My use case is storing largish blobs of structured numeric data (custom machine learning models).

The current solution is json+gzip, which is pretty slow. I was looking at msgpack+zstandard as a drop-in replacement to speed things up and so far the results are very promising.

My benchmark is a set of big dicts, with mostly numpy arrays (which end up as ExtType with large bytes data) and some amount of text/floats. The serialized size is about 300Mb.

One relevant detail, is that even if msgpack itself doesn't massively benefit from streaming, not offering a streaming output means it's impossible to properly leverage the streaming capabilities of the compression library, or to do things like pipelining encoding and I/O.

Adding streaming packing gives about 25-50% speedup to serialization on my dataset. Some of it comes directly from the reduced memory usage in msgpack (avoid a bunch of alloc, repeated memcpy on realloc, less CPU cache trashing, less GC pressure), some probably from being able to leverage streaming compression for further memory gain, and some comes pipelining cpu and I/O.

jfolz:
> My use case is storing largish blobs of structured numeric data (custom machine learning models).
> My benchmark is a set of big dicts, with mostly numpy arrays (which end up as ExtType with large bytes data) and some amount of text/floats. The serialized size is about 300Mb.
That's probably right within the sweet spot. A few dozen arrays, a couple MBs each and not much else. What could also help with arrays is the ability to pack arbitrary objects within ExtType. [`msgpack-numpy`](https://github.com/lebedov/msgpack-numpy) does a goob job or working around the current limitation, but a proper ExtType for ndarrays/tensors would be nice.

PR #348, opened 2019-02-14
StreamingPacker, take 2

huguesb:
When packing larger objects, the current default behavior
of packing the entire object into a buffer and then writing
it into the provided output stream results in massive slowdowns.

To address this case, introduce a StreamingPacker which takes
a callable, to which bytes are sent as they are produced. This
function will typically be the write() method of a file-like
object, but it can be any callable that accepts bytes.

The pure Python implementation writes directly to this callback.

The C implementation only writes to this function when its internal
buffer fills up, and to flush the remainder of the buffer at the
end of a pack*() call. Large elements may also bypass the buffer
completely.

Packer is redefined as a thin wrapper around StreamingPacker to
avoid code duplication while keeping StreamingPacker as simple
as possible.

Add a few more tests to cover large values

### DIFF
--- a/msgpack/__init__.py
+++ b/msgpack/__init__.py
@@ -19,12 +19,115 @@ class ExtType(namedtuple('ExtType', 'code data')):
 
 import os
 if os.environ.get('MSGPACK_PUREPYTHON'):
-    from msgpack.fallback import Packer, unpackb, Unpacker
+    from msgpack.fallback import StreamingPacker, unpackb, Unpacker
 else:
     try:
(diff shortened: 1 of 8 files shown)


methane:
This looks much worse than #347.

methane:
Please benchmark before sending such pull request.

In this time, 6 cases are fundamental: (StreamingPacker, Packer) * (Massive small objects, Some huge objects, some small objects)

This pull request may make (Packer, some small objects) case very slower.
Try microbench for `packb(42)`.

Issue #356, opened 2019-05-10
RFC: use relative imports instead of absolute?


FelixSchwarz:
Currently msgpack uses absolute imports for its modules, e.g.:

    if os.environ.get('MSGPACK_PUREPYTHON'):
        from msgpack.fallback import Packer, unpackb, Unpacker
    else:
        try:
            from msgpack._cmsgpack import Packer, unpackb, Unpacker
        except ImportError:
            from msgpack.fallback import Packer, unpackb, Unpacker

Recently I worked on bundling msgpack inside another project to work around an incompatible system-wide install of msgpack (Fedora 30, borgbackup requires msgpack 0.5.6 but F30 ships 0.6.1).

One thing which made bundling a lot easier was to use relative imports in msgpack:

    if os.environ.get('MSGPACK_PUREPYTHON'):
        from .fallback import Packer, unpackb, Unpacker
    else:
        try:
            from ._cmsgpack import Packer, unpackb, Unpacker
        except ImportError:
            from .fallback import Packer, unpackb, Unpacker

That way I could copy msgpack's code to another location and it would not use the system-wide install.

I could create a PR if you would be interesting in merging this. Otherwise I'll carry some local patches for Fedora until we can drop the bundling.

methane:
I'm OK for it.

FelixSchwarz:
Ok, thanks. I added a pull request.

PR #357, opened 2019-05-10
use relative imports

FelixSchwarz:
This should not change anything but makes it easier to bundle msgpack in other projects (where bundling involves copying msgpack's source code and exposing it under a different Python module name).

Some applications use msgpack to store persistent data and require a specific msgpack version (e.g. borgbackup). Bundling helps in case there is an (incompatible) version of msgpack in a system-wide install.

More background for this change in #356


FelixSchwarz:
Thanks :-)

PR #369, opened 2019-09-18
Use Py_SIZE() for tuple and list

methane:


msgpack/_packer.pyx
```diff
@@ -234,7 +234,7 @@ cdef class Packer(object):
                         ret = self._pack(v, nest_limit-1)
                         if ret != 0: break
             elif not strict_types and PyDict_Check(o):
-                L = len(o)
+                L = Py_SIZE(o)
```
methane:
```suggestion
                L = len(o)
```

o.items() may be overriden so we shouldn't assume that `len(o.items()) == Py_SIZE(o)`.

PR #373, opened 2019-11-13
memoize stringly keys of objects

cfbolz:
When unpacking msgpack objects, store the keys that appear into a memo
dictionary to make them unique. This is useful, because for most sizable
msgpack files the same keys appear again and again, since many objects
have the same "shape" (set of keys). A similar optimization is done in
most json deserializers, eg in CPython:
https://github.com/python/cpython/blob/d89cea15ad37e873003fc74ec2c77660ab620b00/Modules/_json.c#L717

My totally unscientific results: I tried this on two big msgpack files,
a wikidata dump (92 MiB) and a dump of reddit comments (596 MiB). I am
reporting time spent deserializing and memory use of the resulting data
structure. I've included json deserialization numbers as a comparison.
The results I get on my old-ish laptop are:

```
wikidata
                      time   memory
CPython 3.7.5 before  3.42s  1279 MiB
CPython 3.7.5 after   3.43s   883 MiB
PyPy3 7.2 before      6.44s  1380 MiB
PyPy3 7.2 after       4.98s   965 MiB

CPython 3.7.5 json    4.13s   887 MiB
PyPy3 7.2 json        3.54s   958 MiB

reddit
CPython 3.7.5 before   5.62s  3412 MiB
CPython 3.7.5 after    5.20s  1754 MiB
PyPy3 7.2 before      14.72s  3782 MiB
PyPy3 7.2 after        8.37s  2086 MiB

CPython 3.7.5 json     8.64s  1753 MiB
PyPy3 7.2 json        10.52s  2052 MiB
```

For wikidata, there is only a memory improvement on CPython, the time
stays the same. For all other three variants (all of reddit, pypy on
wikidata) both time and memory improve significantly. The reason for the
memory improvements are due to the memoizing, time improves due to
better cache locality due to the smaller working set, and less time
spent in GC in the case of PyPy.

### DIFF
--- a/msgpack/_unpacker.pyx
+++ b/msgpack/_unpacker.pyx
@@ -38,6 +38,7 @@ cdef extern from "unpack.h":
         Py_ssize_t max_array_len
         Py_ssize_t max_map_len
         Py_ssize_t max_ext_len
+        PyObject* memo
 
     ctypedef struct unpack_context:
         msgpack_user user
@@ -61,7 +62,7 @@ cdef inline init_ctx(unpack_context *ctx,
                      const char* encoding, const char* unicode_errors,
                      Py_ssize_t max_str_len, Py_ssize_t max_bin_len,
                      Py_ssize_t max_array_len, Py_ssize_t max_map_len,
-                     Py_ssize_t max_ext_len):
+                     Py_ssize_t max_ext_len, object d):
     unpack_init(ctx)
     ctx.user.use_list = use_list
     ctx.user.raw = raw
@@ -101,6 +102,9 @@ cdef inline init_ctx(unpack_context *ctx,
 
     ctx.user.encoding = encoding
     ctx.user.unicode_errors = unicode_errors
+    Py_INCREF(d)
+    ctx.user.memo = <PyObject*>d
+
 
 def default_read_extended_type(typecode, data):
     raise NotImplementedError("Cannot decode extended type with typecode=%d" % typecode)
@@ -195,9 +199,10 @@ def unpackb(object packed, object object_hook=None, object list_hook=None,
         max_ext_len = buf_len
 
     try:
+        memo = {}
         init_ctx(&ctx, object_hook, object_pairs_hook, list_hook, ext_hook,
                  use_list, raw, strict_map_key, cenc, cerr,
-                 max_str_len, max_bin_len, max_array_len, max_map_len, max_ext_len)
+                 max_str_len, max_bin_len, max_array_len, max_map_len, max_ext_len, memo)
         ret = unpack_construct(&ctx, buf, buf_len, &off)
     finally:
         if new_protocol:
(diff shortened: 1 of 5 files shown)


msgpack/unpack.h
```diff
@@ -193,6 +194,21 @@ static inline int unpack_callback_map_item(unpack_user* u, unsigned int current,
         PyErr_Format(PyExc_ValueError, "%.100s is not allowed for map key", Py_TYPE(k)->tp_name);
         return -1;
     }
+    if (PyUnicode_CheckExact(k) || PyBytes_CheckExact(k)) {
```
jfolz: Should not make a big difference, but this can be combined with the previous condition in l.194 :)

msgpack/_unpacker.pyx
```diff
@@ -101,6 +102,9 @@ cdef inline init_ctx(unpack_context *ctx,
 
     ctx.user.encoding = encoding
     ctx.user.unicode_errors = unicode_errors
+    Py_INCREF(d)
+    ctx.user.memo = <PyObject*>d
```
jfolz: AFAIK `<object>d` will enable refcounting in Cython. Otherwise `object` and `PyObject*` are identical.

methane:
I prefer interning because it speed up string comparison too.

msgpack/unpack.h
cfbolz: Thanks for the review! I fixed this.

msgpack/_unpacker.pyx
cfbolz: However, I didn't manage to follow this suggestion, because the `memo` field is in a C struct, which doesn't support `object` fields.

jfolz:
Isn't interning a bit dangerous, i.e., those strings are there forever until the interpreter exits?

methane:
PyUnicode_InternImmortal() creates immortal string.  It is bit dangerous as you said.
But PyUnicode_InternInplace() doesn't create immortal string.

methane:
BTW, No need to care about Python 2 and bytes object.
Optimize only unicode in Python 3.

jfolz:
Interesting:

> Changed in version 2.3: Interned strings are not immortal

So it's been like that for a very long time and it's just been propagated since. `PyUnicode_InternImmortal` seems to be undocumented - to keep people from using it since it's entirely unnecessary I presume. So interning is exactly as safe and more effective than a custom dict.

cfbolz:
OK, happy to switch to interning in the C version. I would still stick with an explicit dict in the fallback version if that's OK?

jfolz:
Ideally fallback would behave identical to native code. Since you're a PyPy developer I suppose it doesn't play nicely there?

methane:
>  I would still stick with an explicit dict in the fallback version if that's OK?

What's wrong about `sys.intern()`?

cfbolz:
Ok, I investigated, seems PyPy is also good about intern not leaking anything, I wasn't sure :-).

PR #374, opened 2019-11-22
update for Python 3.8

terencehonles:
I believe these changes should be made to support Python 3.8.

I'm not sure if anything else should be changed, but I noticed there were a couple things that weren't updated to reference Python 3.8 in #371. I copied some changes from b8bf3c950c4474d3af82a6b6bda6326b2e197a5e, but I wasn't sure if 3.6 should be removed like 3.5 was in b1d658e7a0aed4e723f9b6b238cc9f2f876e54d7.

From #371 it looks like xenial is being used and 3.8 wouldn't be there so I don't think I can remove the "-dev" suffix and I probably can't add 3.9 via "nightly". Looking the .travis.yml I think(?) 3.8 will be tested via the quay.io/pypa/manylinux1_i686 image.


docker/buildwheel.sh
```diff
@@ -1,10 +1,13 @@
...
+DOCKER_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "$DOCKER_DIR/shared.env"
+
 set -e -x
 
 ARCH=`uname -p`
 echo "arch=$ARCH"
 
-for V in cp37-cp37m cp36-cp36m cp35-cp35m cp27-cp27m cp27-cp27mu; do
+for V in "${PYTHON_VERSIONS[@]}"; do
```
terencehonles: Is there any reason this isn't (and the same for runtests.sh):

```suggestion
for PYBIN in /opt/python/*/bin; do
```

docker/runtests.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
+DOCKER_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "$DOCKER_DIR/shared.env"
+
 set -e -x
 
-for V in cp36-cp36m cp35-cp35m cp27-cp27m cp27-cp27mu; do
+for V in "${PYTHON_VERSIONS[@]}"; do
```
terencehonles: Is there any reason this isn't:

```suggestion
for PYBIN in /opt/python/*/bin; do
```

docker/buildwheel.sh
jfolz: The image ships with lots of unsupported Python versions.

Issue #316, opened 2018-10-22
Add support for Timestamp extension


brandond:
Timestamp extension was added to msgpack spec in https://github.com/msgpack/msgpack/commit/40e3d3db8985528d41a0ded1f9581cede270ec95 over a year ago.
Sample pure python code to serialize/deserialize from datetime was provided in #252, but rejected due to lack of TZ offset support in the timestamp extension.

randallpittman:
I don't understand why it's an issue to require timezone-aware `datetime` for serialization and convert to UTC with `datetime.as_timezone`. And then deserialize to UTC `datetime`. 

If that's making too many decisions for the user then why not just create a UNIX-epoch `(seconds, nanoseconds)` interface a-la `msgpack-c`'s `msgpack_timestamp` and `msgpack_object_to_timestamp` (new as of v3.1.0).

randallpittman:
Furthermore, with the current implementation negative EXT codes are disallowed (since they're reserved for future Msgpack use) so I can't manually pack a timestamp into an `ExtType`. That kind of stinks.

methane:
> but rejected due to lack of TZ offset support in the timestamp extension.

> I don't understand why it's an issue to require timezone-aware datetime for serialization and convert to UTC with datetime.as_timezone. And then deserialize to UTC datetime.

Lack of TZ is not the only reason.  I use it for just an example why msgpack choose unixtime.

datetime is easy (for humans), and unixtime is simple (for machine, program).
msgpack is for simple.  When load / store / transfer time data, unix time should be used.
This is desgin philosophy of msgpack.

I love the design philosophy.  I believe unixtime should be converted only when showing it to humans. Not when deserializing from wire format (=msgpack).

So I don't support datetime *by default*.  This doesn't mean I don't add *option* for datetime in the future, after timestamp support.  But it must not happen before *default*.

methane:
> If that's making too many decisions for the user then why not just create a UNIX-epoch `(seconds, nanoseconds)` interface a-la `msgpack-c`'s `msgpack_timestamp` and `msgpack_object_to_timestamp` (new as of v3.1.0).

"why not"?  Why msgpack-ruby not?  Why msgpack-java not?  (These two are maintained by msgpack
designer)
It's because we're all busy and have more important things.

Python doesn't have specific type for unix timestamp.  It makes difficult to support it.
I need to design API carefully.  It needs lots of time, mental energy, and motivation.
[...]

PR #375, opened 2019-11-22
Add TimestampType and packing/unpacking support

randallpittman:
Here's my attempt at addressing #316. I made a TimestampType that can be packed/unpacked. I did not add any `datetime` conversions, but it will accept a floating-point seconds value for creation and can output to float with `to_float_s`. Also included a test. 

I tried to think it through pretty well and keep it 2/3 compatible. Let me know what you think.

### DIFF
--- a/msgpack/__init__.py
+++ b/msgpack/__init__.py
@@ -5,6 +5,10 @@ from .exceptions import *
 from collections import namedtuple
 from numbers import Number, Integral
 
+import struct
+import sys
+if sys.version_info[0] > 2:
+    long = int
 
 class ExtType(namedtuple('ExtType', 'code data')):
     """ExtType represents ext type in msgpack."""
@@ -44,24 +48,24 @@ class TimestampType(namedtuple('TimestampType', 'seconds, nanoseconds')):
         else:
             # round helps with floating point issues
             nanoseconds = int(round(seconds % 1 * 1e9, 0))
-        seconds = int(seconds // 1)
+        seconds = long(seconds // 1)
         return super(TimestampType, cls).__new__(cls, seconds, nanoseconds)
 
     @staticmethod
     def from_bytes(b):
         """Unpack bytes into a TimestampType."""
         if len(b) == 4:
-            seconds_ = int.from_bytes(b, "big", signed=False)
-            nanoseconds_ = 0
+            seconds = struct.unpack("!L", b)[0]
+            nanoseconds = 0
         elif len(b) == 8:
-            data64 = int.from_bytes(b, "big", signed=False)
-            seconds_ = data64 & 0x00000003ffffffff
-            nanoseconds_ = data64 >> 34
+            data64 = struct.unpack("!Q", b)[0]
+            seconds = data64 & 0x00000003ffffffff
+            nanoseconds = data64 >> 34
         elif len(b) == 12:
-            nanoseconds_, seconds_ = struct.unpack("!Iq", b)
+            nanoseconds, seconds = struct.unpack("!Iq", b)
         else:
             raise ValueError("Timestamp type can only be created from 32, 64, or 96-bit byte objects")
-        return TimestampType(seconds_, nanoseconds_)
+        return TimestampType(seconds, nanoseconds)
 
     def to_bytes(self):
         """Pack TimestampType into bytes."""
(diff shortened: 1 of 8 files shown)


msgpack/_packer.pyx
```diff
@@ -253,6 +254,15 @@ cdef class Packer(object):
                     raise ValueError("EXT data is too large")
                 ret = msgpack_pack_ext(&self.pk, longval, L)
                 ret = msgpack_pack_raw_body(&self.pk, rawval, L)
+            elif type(o) is TimestampType if strict_types else isinstance(o, TimestampType):
+                # This should also be before Tuple because TimestampType is namedtuple.
+                code = -1
+                data = o.to_bytes()
```
jfolz: This calls lots of Python functions, so it's probably quite slow.

randallpittman:
Ok, sure, I wondered about that. I'll try to get that done in Cython.
Randy
________________________________

msgpack/__init__.py
```diff
@@ -3,7 +3,12 @@
 from .exceptions import *
 
 from collections import namedtuple
+from numbers import Number, Integral
```
methane: Don't use them.  Just use int.

msgpack/__init__.py
```diff
@@ -3,7 +3,12 @@
...
 
+import struct
+import sys
+if sys.version_info[0] > 2:
+    long = int
```
methane: I will drop Python 2 support before adding timestamp support.
So no need to care about Python 2 compatibility.

msgpack/__init__.py
```diff
@@ -12,10 +17,74 @@ def __new__(cls, code, data):
             raise TypeError("code must be int")
         if not isinstance(data, bytes):
             raise TypeError("data must be bytes")
+        if code == -1:
+            return TimestampType.from_bytes(data)
         if not 0 <= code <= 127:
             raise ValueError("code must be 0~127")
         return super(ExtType, cls).__new__(cls, code, data)
 
+class TimestampType(namedtuple('TimestampType', 'seconds, nanoseconds')):
```
methane: Don't use namedtuple.  Just use class with `__slots__`.

msgpack/_packer.pyx
```diff
@@ -253,6 +254,15 @@ cdef class Packer(object):
                     raise ValueError("EXT data is too large")
                 ret = msgpack_pack_ext(&self.pk, longval, L)
                 ret = msgpack_pack_raw_body(&self.pk, rawval, L)
+            elif type(o) is TimestampType if strict_types else isinstance(o, TimestampType):
```
methane: No need to `isinstance`.  strict_types is just a backward compatibility option.
No need to care about supporting subclass of Timestamp.

msgpack/__init__.py
randallpittman: May I ask why not? As a user I find it obnoxious to have to convert NumPy types to pure Python types and the `Number` and `Integral` general types alleviate that.

methane: > May I ask why not? 

### Consistency

We don't use them in normal integers.  Why do we support it in Timestamp?
It doesn't make sense at all.

### Performance

```
$ python3 -m timeit -s 'from numbers import Integral' -- 'isinstance(42, Integral)'
500000 loops, best of 5: 443 nsec per loop
$ python3 -m timeit 'isinstance(42, int)'
5000000 loops, best of 5: 59.9 nsec per loop
```

---

>  As a user I find it obnoxious to have to convert NumPy types to pure Python types

Do you really want to use numpy types in Timestamp????

randallpittman: Thanks, everything you said makes sense.

randallpittman:
OK, I think I've addressed all the comments:
- I've added C packing and unpacking implementations (basically pulled from msgpack-c),
- Removed the Python 2 compat stuff (but still check for python3 in the test so it doesn't fail the CI.)
- TimestampType just expects basic python numeric types
- TimestampType is a simple class with `__slots__`.
- Added docs

Happy to get your thoughts.
-Randy

methane:
Thank you, @randallpittman.  I merged this pull request with some modification.

randallpittman:
Glad to contribute.

PR #386, opened 2019-12-05
Use new msgpack spec by default.

### LINKED ISSUE #99: unpack should decode string types by default
msgpack has native text and binary types, but the default behavior of `msgpack.unpackb`  of the text type is to return the still-encoded bytes, rather than the decoded text (unicode on Python 2, str on Python 3). The msgpack spec [specifies](https://github.com/msgpack/msgpack/blob/master/spec.md#type-system) that strings are utf-8, so it should be a logical default.

Changing the default value for `Unpacker.encoding` from `None` to `utf-8` would introduce the desired behavior, but it would be problematic for anyone relying on the default behavior. So I thought I would ask before opening a PR. Would a pull request with this change be accepted?

methane:
Fixes #99

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -37,36 +37,16 @@ Sadly, this doesn't work for upgrade install.  After `pip install -U msgpack-pyt
 msgpack is removed and `import msgpack` fail.
 
 
-Deprecating encoding option
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Compatibility with old format
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-encoding and unicode_errors options are deprecated.
+You can use ``use_bin_type=False`` option to pack ``bytes``
+object into raw type in old msgpack spec, instead of bin type in new msgpack spec.
 
-In case of packer, use UTF-8 always.  Storing other than UTF-8 is not recommended.
+You can unpack old msgpack formatk using ``raw=True`` option.
+It unpacks str (raw) type in msgpack into Python bytes.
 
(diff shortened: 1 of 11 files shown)


README.rst
```diff
@@ -37,36 +37,16 @@ Sadly, this doesn't work for upgrade install.  After `pip install -U msgpack-pyt
...
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Compatibility with old format
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-encoding and unicode_errors options are deprecated.
+You can use ``use_bin_type=False`` option to pack ``bytes``
+object into raw type in old msgpack spec, instead of bin type in new msgpack spec.
 
-In case of packer, use UTF-8 always.  Storing other than UTF-8 is not recommended.
+You can unpack old msgpack formatk using ``raw=True`` option.
```
ThomasWaldmann: typo

test/test_case.py
```diff
@@ -135,4 +134,4 @@ def test_match():
 
 
 def test_unicode():
-    assert unpackb(packb("foobar"), use_list=1) == b"foobar"
+    assert unpackb(packb(u"foobar"), use_list=1) == u"foobar"
```
ThomasWaldmann: somehow makes more sense now. \o/

ThomasWaldmann: are there enough tests that check all "old msgpack" behaviour as well as all "new msgpack" behaviour (str, bytes, unicode).

esp. that using it the old way still works with the new code, when the right options are given.

borgbackup depends on old format and as people have that in their backup repositories, we can't change it.

PR #394, opened 2019-12-11
Support datetime.

### LINKED ISSUE #252: Support serializing datetime objects
Since timestamps are now a part of the msgpack spec, shouldn't this library support serializing them?

methane:
Fixes #252.

### DIFF
--- a/msgpack/fallback.py
+++ b/msgpack/fallback.py
@@ -1,5 +1,6 @@
 """Fallback pure Python implementation of msgpack"""
 
+from datetime import datetime as _DateTime
 import sys
 import struct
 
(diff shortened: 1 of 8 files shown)


msgpack/fallback.py
```diff
@@ -756,6 +780,12 @@ class Packer(object):
         This is useful when trying to implement accurate serialization
         for python types.
 
+    :param bool datetime:
+        If set to true, datetime with tzinfo is packed into Timestamp type.
+        Note that the tzinfo is stripped in the timestamp.
+        You can get UTC datetime with `timestamp=3` option of the Unapcker.
```
thehesiod: Unpacker

methane: Thanks.  Fixed in master.

PR #404, opened 2020-02-07
Add type stubs.

hoefling:
This PR adds [PEP 561](https://www.python.org/dev/peps/pep-0561/) compatibility to `msgpack`.

I keep the type stubs from this PR for development with `msgpack`. The idea is to make them available to everyone by adding them to the distribution. There are no changes to the `msgpack` code itself, the stub is completely separated from the real code.

If you are against adding type stubs, I'd still be grateful for a review that would improve the stub quality - maybe you have a good advice on how to deal with the remaining `Any` types.


methane:
I'm sorry, but I am not an expert of typing and I can not maintain it.
I don't want to merge I can not maintain. So I don't merge this pull request anytime soon.

I will drop Python 2 support completely in next major version (in 2021 or 2022). I will reconsider this after it.

hoefling:
@methane no worries, I see that someone has created a stubs-only package recently: https://github.com/sbdchd/msgpack-types. So this is not an issue anymore, thus closing.

PR #427, opened 2020-06-04
fix #426, add doc to Unpacker.tell(), add tests for tell

### LINKED ISSUE #426: tell() is incorrect after read_bytes()
Using the default `._cmsgpack` implementation of an `Unpacker`, you will get the wrong answer if you call `tell()` after `read_bytes()`. Here's a simple repro:

```
import msgpack
from io import BytesIO
buf = BytesIO(b'my data')
u = msgpack.Unpacker(buf)
print(u.tell())
u.read_bytes(5)
print(u.tell())
```

After calling `u.read_bytes(5)`, `u.tell()` will still return 0. It looks like the `stream_offset` in `_unpacker.pyx` should be advanced in `read_bytes()`, but is currently only advanced by `_unpack()`.

Note that this works correctly in the `.fallback` implementation, so a quick workaround is to use `import msgpack.fallback as msgpack`, but I assume this is much slower than the C implementation.

jfolz:
Fixes the inconsistency in `Unpacker.tell()` between cmsgpack and fallback reported in #426. Also fixes inconsistency with read size greater remaining bytes in input, add tests for tell and doc tell.

### DIFF
--- a/msgpack/fallback.py
+++ b/msgpack/fallback.py
@@ -365,18 +365,19 @@ class Unpacker(object):
         return self._buffer[self._buff_i :]
 
     def read_bytes(self, n):
-        ret = self._read(n)
+        ret = self._read(n, raise_outofdata=False)
         self._consume()
         return ret
 
-    def _read(self, n):
+    def _read(self, n, raise_outofdata=True):
         # (int) -> bytearray
-        self._reserve(n)
+        self._reserve(n, raise_outofdata=raise_outofdata)
         i = self._buff_i
-        self._buff_i = i + n
-        return self._buffer[i : i + n]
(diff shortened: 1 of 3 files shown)


msgpack/fallback.py
```diff
@@ -404,7 +405,7 @@ def _reserve(self, n):
             self._buffer += read_data
             remain_bytes -= len(read_data)
 
-        if len(self._buffer) < n + self._buff_i:
+        if len(self._buffer) < n + self._buff_i and raise_outofdata:
```
jfolz, line 408: This is slightly messy, but required to enable `read_bytes` with more than number of bytes available in the input, which cmsgpack supports.

methane:
Thank you!

PR #436, opened 2020-07-29
fix: datetime before epoch on windows in cython implementation

yanlend:
Cython implementation still used datetime.from_timestamp method, which does not work on windows.
Update the cython implementation to use utc time and delta
and add a regression test to highlight the issue.

### DIFF
--- a/msgpack/unpack.h
+++ b/msgpack/unpack.h
@@ -341,7 +341,26 @@ static int unpack_callback_ext(unpack_user* u, const char* base, const char* pos
         else if (u->timestamp == 0) {  // Timestamp
             py = PyObject_CallFunction(u->timestamp_t, "(Lk)", ts.tv_sec, ts.tv_nsec);
         }
-        else { // float or datetime
+        else if (u->timestamp == 3) {  // datetime
+            // Calculate datetime using epoch + delta 
+            // due to limitations PyDateTime_FromTimestamp on Windows with negative timestamps
+            PyObject *epoch = PyDateTimeAPI->DateTime_FromDateAndTime(1970, 1, 1, 0, 0, 0, 0, u->utc, PyDateTimeAPI->DateTimeType);
+            if (epoch == NULL) {
+                return -1;
+            }
+
+            PyObject* d = PyDelta_FromDSU(0, ts.tv_sec, ts.tv_nsec / 1000);
+            if (d == NULL) {
+                Py_DECREF(epoch);
+                return -1;
+            }
+
+            py = PyNumber_Add(epoch, d);
+            
+            Py_DECREF(epoch);
+            Py_DECREF(d);
+        }
+        else { // float
             PyObject *a = PyFloat_FromDouble((double)ts.tv_nsec);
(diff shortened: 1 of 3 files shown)


yanlend:
Not sure what is wrong in the C/C++ analysis. Please help me and I will fix it.

msgpack/unpack.h
```diff
@@ -342,32 +342,53 @@ static int unpack_callback_ext(unpack_user* u, const char* base, const char* pos
             py = PyObject_CallFunction(u->timestamp_t, "(Lk)", ts.tv_sec, ts.tv_nsec);
         }
         else { // float or datetime
-            PyObject *a = PyFloat_FromDouble((double)ts.tv_nsec);
-            if (a == NULL) return -1;
+            if (u->timestamp == 3) {  // datetime
```
methane, line 345: Use `else if`, not `else { if`.

msgpack/unpack.h
```diff
@@ -342,32 +342,53 @@ static int unpack_callback_ext(unpack_user* u, const char* base, const char* pos
...
-            Py_DECREF(c);
+                PyObject* d = PyDelta_FromDSU(0, ts.tv_sec, ts.tv_nsec / 1000);
+                if (d == NULL) {
+                    Py_DECREF(t_epoch);
+                    return -1;
+                }
 
-            if (u->timestamp == 3) {  // datetime
-                PyObject *t = PyTuple_Pack(2, a, u->utc);
+                py = PyObject_CallMethod(t_epoch, "__add__", "(O)", d);
```
methane, line 371: Use `PyNumber_Add(t_epoch, d)`

msgpack/unpack.h
```diff
@@ -342,32 +342,53 @@ static int unpack_callback_ext(unpack_user* u, const char* base, const char* pos
...
+                if (t == NULL) {
+                    return -1;
+                }
 
-            PyObject *c = PyLong_FromLongLong(ts.tv_sec);
-            if (c == NULL) {
-                Py_DECREF(b);
-                return -1;
-            }
+                PyObject *t_epoch = PyDateTime_FromTimestamp(t);
```
methane, line 359:
```c
    PyObject *epoch = PyDateTimeAPI->DateTime_FromDateAndTimeAndFold(1970, 1, 1, 0, 0, 0, 0, u->utc, 0, PyDateTimeAPI->DateTimeType);
```

yanlend: Done

msgpack/unpack.h
yanlend: Done

msgpack/unpack.h
yanlend: Done

msgpack/unpack.h
yanlend: It seems the proposed method is not available in Python 3.5

We don't need the "...AndFold" anyways, I could use "PyDateTime_FromDateAndTime"

methane:
Thank you!

PR #443, opened 2020-11-02
Setup Timestamp to support rich comparisons

fried:
When working with Timestamp objects I was surprised to find out they couldn't be compared to each other beyond eq and ne.


methane:
> When working with Timestamp objects

Please don't. mgpack.Timestamp is not for application use. It is just a temporary object between msgpack and application timestamp type.

PR #450, opened 2020-12-16
Change first command to "what you should do"

RichardBronosky:
When skimming the page, it's very easy to misinterpret "don't do `code B` instead do `code A`"


methane:
I don't like this reorganization.

PR #466, opened 2021-02-26
Nicer error when packing a datetime without tzinfo

bem7:
When attempting to pack a `datetime.datetime` object with `msgpack.packb(dt, datetime=True)` where the `dt` object is missing `tzinfo`, the library throws the generic error `TypeError: can not serialize 'datetime.datetime' object`.

That threw me off as I thought the problem related to `datetime=True`, and didn't realise I was missing `tzinfo`. This small patch outputs a more specific error message for that edge case.

### DIFF
--- a/msgpack/_packer.pyx
+++ b/msgpack/_packer.pyx
@@ -274,7 +274,9 @@ cdef class Packer(object):
                 if ret == 0:
                     ret = msgpack_pack_raw_body(&self.pk, <char*>view.buf, L)
                 PyBuffer_Release(&view);
-            elif self.datetime and PyDateTime_CheckExact(o) and datetime_tzinfo(o) is not None:
+            elif self.datetime and PyDateTime_CheckExact(o):
+                if datetime_tzinfo(o) is None:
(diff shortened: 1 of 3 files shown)


msgpack/_packer.pyx
```diff
@@ -274,7 +274,9 @@ cdef class Packer(object):
                 if ret == 0:
                     ret = msgpack_pack_raw_body(&self.pk, <char*>view.buf, L)
                 PyBuffer_Release(&view);
-            elif self.datetime and PyDateTime_CheckExact(o) and datetime_tzinfo(o) is not None:
+            elif self.datetime and PyDateTime_CheckExact(o):
+                if datetime_tzinfo(o) is None:
+                    PyErr_Format(ValueError, b"can not serialize '%.200s' object where tzinfo=None", Py_TYPE(o).tp_name)
```
methane, line 279: This is backward incompatible change.
User can not use `default` function to pack datetime without tzinfo.

bem7:
Addressed your comment in this new commit – it now respects the `default` function.

The logic is getting a little twisted, though. If you don't think it's worth it for such a minor convenience, I understand.

Issue #219, opened 2017-02-08
OSS-Fuzz


methane:
https://github.com/google/oss-fuzz

DaveLak:
@methane is the OSS-Fuzz integration still something this project is interested in? If so, I'd be happy to assist.

I put up a PR on the OSS-Fuzz repo fixing the fuzzer build there (https://github.com/google/oss-fuzz/pull/12442) but if you'd like to see the integration improved I'd be happy to migrate it into this repo and assist with coverage improvements and documentation similar to what I've done for other projects.

Let me know if you would like me to share a more detailed plan for you and other maintainers to consider 🙂

PR #469, opened 2021-03-01
Add hypothesis-powered tests

hexagonrecursion:
This works towards https://github.com/msgpack/msgpack-python/issues/219
See https://google.github.io/oss-fuzz/getting-started/new-project-guide/python-lang/#hypothesis

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -42,7 +42,7 @@ jobs:
 
       - name: Run test (3.8)
         run: |
-          pip install pytest
+          pip install pytest hypothesis
           pip install -v msgpack --only-binary :all: --no-index -f dist/wheelhouse
           pytest -v test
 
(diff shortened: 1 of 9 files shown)


hexagonrecursion:
This is wrong. It should not be that hard to add a new dependency. Why are you repeating the same information in so many places? You have tox.ini. Use it.

methane:
I do not use tox.

hexagonrecursion:
I have finally hit genuine test failures. I'll look into them later.
```
>       assert repr(obj) == repr(got)

E       assert '{b\'"b\': No...641537: None}' == '{2.000000001...\'"b\': None}'

E         - {2.0000000011641537: None, b'"b': None}

E         + {b'"b': None, 2.0000000011641537: None}
```

methane:
> I have finally hit genuine test failures. I'll look into them later.

Dicts in old Pythons don't preserve order.

methane:
I'm not interested in fuzzing packer. I'm interested in Unpacker because it can be attack surface.

PR #473, opened 2021-03-18
cimport uint64_t instead of using ctypedef

vladima:
Cython already has its out [internal definition](https://github.com/cython/cython/blob/0.29.x/Cython/Includes/libc/stdint.pxd#L16) of `uint64_t` so it might as well  be used here. In addition `ctypedef` causes a build break if `msgpack` is compiled with Cython3.


methane:
thank you.

PR #499, opened 2022-03-11
Use PyFloat_Pack8() on Python 3.11a7

vstinner:
Python 3.11a7 adds public functions:

* PyFloat_Pack4(), PyFloat_Pack8()
* PyFloat_Unpack4(), PyFloat_Unpack8()

https://bugs.python.org/issue46906

### DIFF
--- a/msgpack/pack_template.h
+++ b/msgpack/pack_template.h
@@ -568,7 +568,12 @@ static inline int msgpack_pack_float(msgpack_packer* x, float d)
 {
     unsigned char buf[5];
     buf[0] = 0xca;
+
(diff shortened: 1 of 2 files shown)


vstinner:
msgpack currently fails to build on Python 3.11 since the _PyFloat_Pack4() function moved to the internal C API.

I just added a new clean public C API for msgpack: see [bpo-46906]( https://bugs.python.org/issue46906)). This PR uses it.

Python 3.11 alpha7 is not released yet. It's scheduled in one month (at 2022-04-05): https://peps.python.org/pep-0664/

PR #504, opened 2022-04-26
Do not fail silently when unpacking large messages

### LINKED ISSUE #503: Default max_buffer_size causes silent failures when streaming
Thanks for a useful library!

Consider the following repro:
```
import os
import msgpack
import tempfile

original_values = [
    (b"x" * 1024) * (1024 * 40 * i)
    for i in range(1, 5)
]

with tempfile.TemporaryDirectory() as tmpdir:
    fn = os.path.join(tmpdir, "test.msgpack")
    with open(fn, "wb") as f:
        for v in original_values:
            msgpack.pack(v, f)
    
    with open(fn, "rb") as f:
        unpack = msgpack.Unpacker(f, max_buffer_size=0)
        values = list(unpack)
    assert values == original_values

    with open(fn, "rb") as f:
        unpack = msgpack.Unpacker(f)
        values = list(unpack)
    # failed to unpack some of the stream, but we wouldn't know if we didn't have original_values to check against
    assert values == original_values
```

hauntsaninja:
Fixes #503

Raising ValueError matches the behaviour of the pure Python version.
I'm not sure that this is the best place to raise this message; it might
make more sense when reading headers or in `append_buffer` or
something. It's also possible that we want to raise BufferFull instead
of ValueError.

### DIFF
--- a/msgpack/_unpacker.pyx
+++ b/msgpack/_unpacker.pyx
@@ -441,12 +441,17 @@ cdef class Unpacker(object):
         self.buf_tail = tail + _buf_len
 
     cdef read_from_file(self):
+        current_size = self.buf_tail - self.buf_head
         next_bytes = self.file_like_read(
-                min(self.read_size,
-                    self.max_buffer_size - (self.buf_tail - self.buf_head)
-                    ))
+            min(self.read_size, self.max_buffer_size - current_size + 1)
+        )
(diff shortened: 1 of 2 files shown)


hauntsaninja:
https://github.com/msgpack/msgpack-python/pull/505 to fix the CI failure

PR #513, opened 2022-08-01
Fallback to compiler defines when __BYTE_ORDER is not available

kulikjak:
On some systems (I tested it on Oracle Solaris and FreeBSD),  `__BYTE_ORDER`, `__LITTLE_ENDIAN` and `__BIG_ENDIAN` from `arpa/inet.h` are not available, which can cause failures on big endian systems (because little endian is the default). This adds a fallback to common compiler defines in those cases.

Addresses issues raised in #495.

### DIFF
--- a/msgpack/sysdep.h
+++ b/msgpack/sysdep.h
@@ -61,8 +61,6 @@ typedef unsigned int _msgpack_atomic_counter_t;
 #endif
 #endif
 
-#else
-#include <arpa/inet.h>  /* __BYTE_ORDER */
 #endif
 
 #if !defined(__LITTLE_ENDIAN__) && !defined(__BIG_ENDIAN__)
--- a/msgpack/sysdep.h
+++ b/msgpack/sysdep.h
@@ -66,6 +66,15 @@ typedef unsigned int _msgpack_atomic_counter_t;
 #endif
 
 #if !defined(__LITTLE_ENDIAN__) && !defined(__BIG_ENDIAN__)
+/*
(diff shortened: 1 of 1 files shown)


methane:
After how `__BYTE_ORDER` and `__BYTE_ORDER__` are used in world, I prefer using just `__BYTE_ORDER__`.
Please make it simple.

kulikjak:
Done, thanks!

methane:
Remove this too:

```
#else
#include <arpa/inet.h>  /* __BYTE_ORDER */
```

klemensn:
msgpack-python 1.0.4 regression tests pass 100% with and without this fix on OpenBSD/sparc64.
Here is an example test run of 1.0.4 without this fix: https://gist.github.com/936496c1bea568a0f741ae9b3ec54d39

Passing tests but failing runtime consumers at the same time suggests a lack of endianess coverage.

klemensn:
> Remove this too:
> 
> ```
> #else
> #include <arpa/inet.h>  /* __BYTE_ORDER */
> ```

Can you elaborate why?

This include is required by [ntohs(3)](https://man.openbsd.org/ntohs.3) later in the same file:
https://github.com/msgpack/msgpack-python/pull/513/files#diff-5c123c916851553e85829468dc2fadee6b1ed6edda136c8d544d5a0eaf594d99R90

Accidentially, it seems, this builds fine on OpenBSD, but breaks at least in Ubuntu 20.04, see
https://github.com/borgbackup/borg/pull/7181#issuecomment-1336118006 and https://github.com/borgbackup/borg/pull/7181/commits/e79986b2835291431f04683423304a02b039c0c1.

methane:
#514

PR #519, opened 2022-09-11
Drop python2 support

### LINKED ISSUE #518: Drop Python 2 support
According to [pypi page](https://pypi.org/project/msgpack/), the python2 support has been dropped with release 1.0.
The main branch in the repository is currently v.1.0.4 release (https://github.com/msgpack/msgpack-python/blob/main/msgpack/__init__.py#L9).

However, there is python2 specific code in the repository. For example:

 - `setup.py` checks for python2
 - `msgpack/fallback.py` could be simplified
 - python2 is checked in `tox.ini`
 - `test/test_timestamp.py` skip some tests with python2 (`@pytest.mark.skipif(sys.version_info[0] == 2`) ; `test/test_extension.py` has also a conditional

Are you interested by a PR to remove them?

sblondon:
The PR removes python2 references and cases.
`tox` command is green on cpython3.9 (py39-c and py39-pure) on my local machine.

Close #518

### DIFF
--- a/msgpack/ext.py
+++ b/msgpack/ext.py
@@ -164,8 +164,6 @@ class Timestamp(object):
     def to_datetime(self):
         """Get the timestamp as a UTC datetime.
 
-        Python 2 is not supported.
-
         :rtype: datetime.
         """
         utc = datetime.timezone.utc
@@ -176,8 +174,6 @@ class Timestamp(object):
     @staticmethod
     def from_datetime(dt):
         """Create a Timestamp from datetime with tzinfo.
-
-        Python 2 is not supported.
 
         :rtype: Timestamp
         """
--- a/msgpack/ext.py
+++ b/msgpack/ext.py
@@ -5,7 +5,6 @@ import sys
 import struct
 
 
-int_types = int
 try:
     _utc = datetime.timezone.utc
 except AttributeError:
@@ -49,9 +48,9 @@ class Timestamp(object):
 
         Note: Negative times (before the UNIX epoch) are represented as negative seconds + positive ns.
         """
-        if not isinstance(seconds, int_types):
+        if not isinstance(seconds, int):
             raise TypeError("seconds must be an interger")
-        if not isinstance(nanoseconds, int_types):
+        if not isinstance(nanoseconds, int):
             raise TypeError("nanoseconds must be an integer")
         if not (0 <= nanoseconds < 10**9):
             raise ValueError(
--- a/msgpack/ext.py
+++ b/msgpack/ext.py
@@ -5,9 +5,6 @@ import sys
 import struct
 
 
-_utc = datetime.timezone.utc
-
-
 class ExtType(namedtuple("ExtType", "code data")):
     """ExtType represents ext type in msgpack."""
 
--- a/msgpack/ext.py
+++ b/msgpack/ext.py
@@ -168,7 +168,8 @@ class Timestamp(object):
 
         :rtype: datetime.
         """
-        return datetime.datetime.fromtimestamp(0, _utc) + datetime.timedelta(
+        utc = datetime.timezone.utc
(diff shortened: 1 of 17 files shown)


msgpack/ext.py
```diff
@@ -5,17 +5,10 @@
...
-    int_types = (int, long)
-    _utc = None
-else:
-    int_types = int
-    try:
-        _utc = datetime.timezone.utc
-    except AttributeError:
-        _utc = datetime.timezone(datetime.timedelta(0))
+try:
+    _utc = datetime.timezone.utc
```
ThomasWaldmann, line 9: This is present since at least python 3.4:

https://docs.python.org/3.4/library/datetime.html#datetime.timezone.utc

msgpack/ext.py
```diff
@@ -55,9 +48,9 @@ def __init__(self, seconds, nanoseconds=0):
 
         Note: Negative times (before the UNIX epoch) are represented as negative seconds + positive ns.
         """
-        if not isinstance(seconds, int_types):
+        if not isinstance(seconds, int):
             raise TypeError("seconds must be an interger")
```
ThomasWaldmann, line 52: there is a typo.

msgpack/ext.py
sblondon: I pushed a new commit to remove the try/except clause.

msgpack/ext.py
sblondon: I don't see the typo. What is it?
In case you talk about the change `int_types` to `int`, `int_types` was a variable needed previously but its value is always `int` with python3 so `int_types` is removed by this PR.

ThomasWaldmann: interger

msgpack/ext.py
```diff
@@ -5,10 +5,7 @@
 import struct
 
 
-try:
-    _utc = datetime.timezone.utc
-except AttributeError:
-    _utc = datetime.timezone(datetime.timedelta(0))
+_utc = datetime.timezone.utc
```
ThomasWaldmann, line 8: Maybe using it directly at all places where it is needed would be prettier.

msgpack/ext.py
sblondon: It's fixed.

msgpack/ext.py
sblondon: I inlined the `_utc` in function and tests, under `utc` name.

msgpack/ext.py
```diff
@@ -5,17 +5,7 @@
...
-if PY2:
-    int_types = (int, long)
-    _utc = None
-else:
-    int_types = int
-    try:
-        _utc = datetime.timezone.utc
-    except AttributeError:
-        _utc = datetime.timezone(datetime.timedelta(0))
+_utc = datetime.timezone.utc
```
methane, line 8: Could you remove this?

sblondon: Oops, I forgot to remove the line. It's done now.

sblondon:
The last commit should fix the `black` step error in the Github CI (https://github.com/msgpack/msgpack-python/actions/runs/3630468029/jobs/6156022655).

However, when I run `black .` on my system, there are several other changes (a lot of `a**b` modified to `a ** b`). They are not added to the commit in order to limit the changes. I can add them if you want.

ThomasWaldmann:
@sblondon it might depend on the black version used.

At @borgbackup, we have a version-fixed requirement in `requirements.d/codestyle.txt` so everybody uses exactly the same version.

sblondon:
> At @msgpack there is also such a definition in `requirements.txt`, did you use it?

No, I runned the `black` which is installed on the system, not the version defined in `requirements.txt` file. So the differences generated by `black` are not useful here: all checks (including black) are green on Github CI.

PR #527, opened 2023-01-18
fallback: Fix packing multidim memoryview

methane:


pohlt:
LGTM

Thanks a lot for the quick fix!

PR #536, opened 2023-04-01
build pypy wheels in CI

dand-oss:
build pypy wheels in CI


methane:
Why? Do you confirmed that extension module boost PyPy performance?
When I tried, binary extension slows PyPy greatly.

PR #552, opened 2023-08-18
Be friendlier with type cheker

senyai:
When using a compiled version of the extension, the type checker couldn't resolve anything. Now it can. And with this PR it's easier to navigate code from vscode.

### DIFF
--- a/msgpack/__init__.py
+++ b/msgpack/__init__.py
@@ -3,13 +3,14 @@ from .ext import ExtType, Timestamp
 
 import os
 import sys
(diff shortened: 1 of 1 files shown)


methane:
I don't want to import typing module.

senyai:
@methane, can you elaborate on why you don't want to import `typing` module? I personally think that types reduce debugging time significantly.

methane:
Importing typing takes significant time and RAM.
I don't want to force all users to pay the cost.

PR #557, opened 2023-09-05
prepare for 1.0.6

methane:


ChangeLog.rst
```diff
@@ -1,3 +1,12 @@
+1.0.6rc1
+========
+
+Release Date: 2023-09-13
+
+* Add Python 3.12 wheels (#517)
+* Remove Python 2.7, 3.6, and 3.7 support
```
ThomasWaldmann, line 7: So, py27 is now officially unsupported?

I noticed a while ago that some docs still talk about py2.
Maybe there is also some code still relating to py2?
Shall I make a PR to remove that?

methane: Yes. I can not maintain Python 2.7 support.
Running/debugging on Python 2.7 is difficult today. So I will reject any Python 2 issues.
For example, https://github.com/actions/setup-python/issues/672

Code cleanup for removing Python 2 support will be checked into msgpack 1.1.
For msgpack 1.0.6, I want to release RC as fast as possible.

I will update README/doc between 1.0.6rc and 1.0.6. PRs are welcome.

ThomasWaldmann: Yeah, a 1.0.6 soon would be very welcome!

PR #560, opened 2023-09-07
fix .readthedocs.yaml, fixes #559

### LINKED ISSUE #559: fix .readthedocs.yaml
The api docs look rather empty now, guess that "basic" config is not good enough for msgpack docs:

https://readthedocs.org/api/v2/build/21850585.txt

ThomasWaldmann:
currently, api docs seem rather empty.

try:

- apt install build-essential (so we have gcc, make, etc.)
- install from requirements.txt
- "make cython"
- "pip install ."

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -4,9 +4,20 @@
 version: 2
 
 build:
-  os: ubuntu-22.04
-  tools:
-    python: "3.11"
+    os: ubuntu-22.04
+    tools:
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
Let's see if this works after it is merged...

ThomasWaldmann:
@methane can you merge please so we can see if this fixes the issue?

ThomasWaldmann:
@methane ping?

PR #598, opened 2024-05-03
use ruff instead of black

methane:


ThomasWaldmann:
@methane I'm curious: can you tell why you switched from black to ruff?

methane:
Ruff is fast and provide black + flake8 + pyupgrade in single binary.
I am migrating all my Python projects into ruff for reduce my cognitive load.

PR #620, opened 2024-08-20
MNT: use PyLong_* instead of PyInt_*

tacaswell:
9af421163cb8081414be347038dee7a82b29e8dd in Cython (not released yet) removed back-compatibility `#define`.

This was the most obvious fix, but I'm not sure how far back it works.  I can also put the required `#define` (conditionally) at the top of this file if you would prefer that.


tacaswell:
I appreciate the irony of opening this right after claiming cython rarely breaks things 😞 .

PR #637, opened 2025-05-30
update Cython to 3.1.1

methane:


ThomasWaldmann:
Recent setuptools releases require Python >= 3.9.

https://setuptools.pypa.io/en/latest/history.html#v75-4-0

methane:
Recent setuptools emits warning about `license = {text="Apache 2.0"}` when building from source.
But I decided to ignore it for now. Next msgpack version (v1.1.1) will ship with Cython 3.1.1 and Python 3.8 support.

Issue #613, opened 2024-06-20
Build fails on 3.14 no-GIL


clin1234:
```
$ make
cython msgpack/_cmsgpack.pyx
python setup.py build_ext -i -f
running build_ext
building 'msgpack._cmsgpack' extension
creating build
creating build/temp.linux-x86_64-cpython-314
creating build/temp.linux-x86_64-cpython-314/msgpack
x86_64-linux-gnu-gcc -fno-strict-overflow -Wsign-compare -DNDEBUG -g -O2 -Wall -g -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -flto=auto -ffat-lto-objects -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -fPIC -I. -I/tmp/venv/include -I/usr/include/python3.14t -c msgpack/_cmsgpack.c -o build/temp.linux-x86_64-cpython-314/msgpack/_cmsgpack.o
msgpack/_cmsgpack.c:2343:80: error: unknown type name ‘__pyx_vectorcallfunc’; did you mean ‘vectorcallfunc’?
 2343 | ject *__Pyx_PyVectorcall_FastCallDict(PyObject *func, __pyx_vectorcallfunc vc, PyObject *const *args, size_t nargs, PyObject *kw);
      |                                                       ^~~~~~~~~~~~~~~~~~~~
      |                                                       vectorcallfunc
msgpack/_cmsgpack.c: In function ‘__pyx_pf_7msgpack_9_cmsgpack_6Packer_2__dealloc__’:
msgpack/_cmsgpack.c:6941:3: warning: ‘Py_OptimizeFlag’ is deprecated [-Wdeprecated-declarations]
 6941 |   if (unlikely(__pyx_assertions_enabled())) {
      |   ^~
In file included from /usr/include/python3.14t/Python.h:72,
                 from msgpack/_cmsgpack.c:16:
/usr/include/python3.14t/cpython/pydebug.h:13:37: note: declared here
   13 | Py_DEPRECATED(3.12) PyAPI_DATA(int) Py_OptimizeFlag;
      |                                     ^~~~~~~~~~~~~~~
msgpack/_cmsgpack.c: At top level:
msgpack/_cmsgpack.c:21382:69: error: unknown type name ‘__pyx_vectorcallfunc’; did you mean ‘vectorcallfunc’?
[...]

PR #641, opened 2025-07-20
Add no-GIL interpreter support

clin1234:


.github/workflows/wheel.yml
```diff
@@ -34,6 +34,7 @@ jobs:
           CIBW_TEST_REQUIRES: "pytest"
           CIBW_TEST_COMMAND: "pytest {package}/test"
           CIBW_SKIP: "pp* cp38-macosx_*"
+          CIBW_ENABLE: cpython-freerelease
```
lysnikolaou, line 37:
```suggestion
          CIBW_ENABLE: cpython-freethreading
```

test/test_multithreading.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
```
lysnikolaou, line 1: According to https://github.com/msgpack/msgpack-python/issues/613#issuecomment-2914866092, sharing `Packer` objects wouldn't be supported under free-threading, so this test is probably not needed. Could we run the test suite with `pytest-run-parallel` instead?

lysnikolaou (commented):
Hey @clin1234! Thanks for working on this. I'm not the maintainer here, but I'm leaving a couple of comments to help move this forward.

test/test_multithreading.py
clin1234: Whatever works: I borrowed this test from another fork, and I'm still learning how to modernize existing `pytest` tests to use `pytest-run-parallel`

lysnikolaou: What I would do is install `pytest-run-parallel` in CI and then add a step in the main test job that looks something like this:

```yml
- name: Install pytest-run-parallel under free-threading
  if: matrix.py == '3.14t-dev' || matrix.py == '3.13t'
  shell: bash
  run: |
    pip install pytest-run-parallel

- name: Run tests in parallel under free-threading
  if: matrix.py == '3.14t-dev' || matrix.py == '3.13t'
  shell: bash
  run: |
    pytest -v --parallel-threads=auto --iterations=20 test
```

lysnikolaou (commented):
Thanks for iterating on this so quickly @clin1234! Left some more comments on here. Hope these are helpful!

test/test_buffer.py
```diff
@@ -17,7 +17,7 @@ def test_unpack_bytearray():
     obj = unpackb(buf, use_list=1)
     assert [b"foo", b"bar"] == obj
     expected_type = bytes
-    assert all(type(s) == expected_type for s in obj)
+    assert all(type(s) is expected_type for s in obj)
```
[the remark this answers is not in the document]
clin1234: Fixed lints as well in this PR

test/test_multithreading.py
lysnikolaou: Great! I think this file can go now, since we're testing with `pytest-run-parallel` and sharing `Packer` objects is not supported.

.github/workflows/test.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest", "windows-latest", "macos-latest"]
-        py: ["3.14-dev", "3.13", "3.12", "3.11", "3.10", "3.9"]
+        py: ["3.14-dev", "3.14t-dev", "3.13", "3.13t", "3.12", "3.11", "3.10", "3.9"]
```
khneal, line 13: If "3.13t" takes ANY additional effort, consider leaving it behind and focusing on "3.14t".
[After all](https://docs.python.org/3.14/howto/free-threading-python.html)... "free-threaded mode is experimental" in 3.13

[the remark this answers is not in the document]
clin1234: Will do.

clin1234: ?

ferdnyc: I think @khneal was suggesting to just focus on only 3.14t instead of 3.13t, if supporting both involves more work — but I don't get the impression it does. The same configuration applies to both, and the CI system is the only thing that has to care about whether there's one freethreading test version or two.

PR #647, opened 2025-10-05
Enable preleases in wheel.yml

### LINKED ISSUE #644: Request for wheels for Python 3.14rc2 for testing purposes
*   **Request**: We request the creation of `msgpack-1.1.1-cp314-*` wheels for Python 3.14.
*   **Source**: https://pypi.org/project/msgpack/#files
*   **Rationale**: Providing these wheels will allow developers to test their applications against the same binaries that will be distributed to end-users upon the official release of Python 3.14.
*   **Alternative**: While it is possible for us to create our own builds, they would likely differ from the official releases. This would undermine the validity of our testing, as we would not be testing against the exact files that will be publicly available.

mborsetti:
Enables the automated building of wheel pre-releases needed by developers for testing purposes.

Closes #644 and #646

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -31,6 +31,7 @@ jobs:
       - name: Build
         uses: pypa/cibuildwheel@v3.1.1
         env:
+          CIBW_PRERELEASE_PYTHONS: "auto"
           CIBW_TEST_REQUIRES: "pytest"
           CIBW_TEST_COMMAND: "pytest {package}/test"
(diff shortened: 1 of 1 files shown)


methane:
it is not required to build wheels for RC version.
Wheels for Python 3.14 are already built in our CI.

methane:
its pypi, not ci.
You can find Python 3.14 wheels at here. https://github.com/msgpack/msgpack-python/actions/runs/18027713595

PR #657, opened 2025-10-24
Update README with thread safety section

lysnikolaou:
Since the next version is going to be supporting the free-threaded build, I thought it might be nice to have a note about it in the README. If it's desirable, I can also update the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,41 @@ However, a tuple is lighter than a list.
 You can use `use_list=False` while unpacking when performance is important.
 
 
+### Thread safety (CPython 3.14+ free-threaded build)
+
+msgpack is designed to be thread-safe when used with CPython 3.14's
+free-threaded build (PEP 703).
+
+#### Thread-Safety Guarantees
+
+* **Individual `Packer` and `Unpacker` instances are thread-safe**:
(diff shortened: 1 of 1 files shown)


lysnikolaou:
@methane A gentle ping on this. I don't think the Copilot comments are very relevant and I still feel this PR is ready to go.

clin1234:
> @methane A gentle ping on this. I don't think the Copilot comments are very relevant and I still feel this PR is ready to go.

Your wish is about to be fulfilled: https://github.com/msgpack/msgpack-python/releases/tag/v1.2.0
;-)

README.md
```diff
@@ -204,6 +204,41 @@ However, a tuple is lighter than a list.
...
 
 
+### Thread safety (CPython 3.14+ free-threaded build)
+
+msgpack is designed to be thread-safe when used with CPython 3.14's
+free-threaded build (PEP 703).
+
+#### Thread-Safety Guarantees
+
+* **Individual `Packer` and `Unpacker` instances are thread-safe**:
```
methane, line 214: thread-safe is a very strong guarantee. Packer and Unpacker are designed to not crash when used from multiple threads, but the correctness of the output depends on the types and timing of the methods being called simultaneously.

PR #659, opened 2025-12-10
Have unpacked datetime timezone being local

cielavenir:
In some cases, timezone being local can be better than utc.

Firstly, may I ask if having such option is worth?

### DIFF
--- a/msgpack/ext.py
+++ b/msgpack/ext.py
@@ -157,9 +157,10 @@ class Timestamp:
         :rtype: `datetime.datetime`
         """
         utc = datetime.timezone.utc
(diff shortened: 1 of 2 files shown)


cielavenir:
Hm, at least this cannot be default behavior as Windows does not support it

methane:
> Firstly, may I ask if having such option is worth?

No.

cielavenir:
@methane Do you mind elaborating? Do you mean msgpack-python should return UTC and converting to local timezone should be done by the user?

methane:
> Do you mean msgpack-python should return UTC and converting to local timezone should be done by the user?

Yes.

PR #660, opened 2026-01-02
Add typing in Python code

### LINKED ISSUE #448: Adding type stubs
In the course of using msgpack in a project I've written some basic type stubs for the parts of the API I'm using.
I don't think it would be too hard to fill out the type stubs for the rest of the API.

Would you accept a PR adding stub files to the package?
Not sure if you have a preference on adding the types to the code itself or keeping them as separate stub files.

related: https://www.python.org/dev/peps/pep-0561/

peasoft:
Add Python 3.10-style typing for msgpack-python.

Compared to [sbdchd/msgpack-types](https://github.com/sbdchd/msgpack-types) and #404, this PR added typing for all Python codes.

Closes #448

### DIFF
--- a/msgpack/ext.py
+++ b/msgpack/ext.py
@@ -6,6 +6,7 @@ from collections import namedtuple
 
 class ExtType(namedtuple("ExtType", "code data")):
     """ExtType represents ext type in msgpack."""
+
     code: int
     data: bytes
 
@@ -16,7 +17,7 @@ class ExtType(namedtuple("ExtType", "code data")):
             raise TypeError("data must be bytes")
         if not 0 <= code <= 127:
             raise ValueError("code must be 0~127")
-        return super().__new__(cls, code, data) # type: ignore
+        return super().__new__(cls, code, data)  # type: ignore
 
 
 class Timestamp:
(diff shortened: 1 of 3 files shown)


msgpack/ext.py
```diff
@@ -28,7 +32,7 @@ class Timestamp:
 
     __slots__ = ["seconds", "nanoseconds"]
 
-    def __init__(self, seconds, nanoseconds=0):
+    def __init__(self, seconds: int, nanoseconds=0):
```
KowalskiThomas, line 35: Shouldn't this also have typing for `nanoseconds`?  Or are you deliberately not adding type annotations where the type can be deduced maybe? 

I'm asking because I believe e.g. `mypy` wouldn't like this (partial type hinting in function definitions)

peasoft: My typing is intended for VS Code's (and maybe other editors') auto complete, so I didn't add type annotations where the type can be deduced.

KowalskiThomas: FWIW if you're planning to get this merged, I'd highly recommend to go ahead and fully type it, even for things that pyright (which VSCode uses) can infer -- it'll probably be helpful to more people and is only marginally more work!

methane:
@copilot resolve the merge conflicts in this pull request

PR #661, opened 2026-02-13
Add a 'Bypass' type that just concatenates its contents to output

marcoffee:
This is a very useful feature for when you have a class that know how to convert its structure to msgpack (specially from compiled modules) and you don't want to have its contents converted to the binary object by returning the converted msgpack from `default`.

As an example, `orjson` has a similar feature, [which they call `Fragment`](https://github.com/ijl/orjson/blob/ec2b066cae79ae4a90ed126ac5723335dd99e408/README.md#fragment).

Some use cases include, but are not limited to:
- bypassing valid msgpack data from compiled classes (specially deep classes, with many subclasses, dicts, lists)
- caching msgpack output when you need to serialize a class multiple times on the same or different calls
- appending an already serialized payload to a new payload (such as including it on a list or as a value on a dict)

PS: Sorry for closing and reopening it, I had some failed linter tests and wanted to fix them, but ended up pushing a version without any changes, which triggered GitHub's auto close feature.

### DIFF
--- a/msgpack/__init__.py
+++ b/msgpack/__init__.py
@@ -2,7 +2,7 @@
 import os
 
 from .exceptions import *  # noqa: F403
-from .ext import ExtType, Timestamp
+from .ext import Bypass, ExtType, Timestamp
 
 version = (1, 1, 2)
 __version__ = "1.1.2"
--- a/msgpack/__init__.py
+++ b/msgpack/__init__.py
@@ -2,7 +2,7 @@
 import os
 
 from .exceptions import *  # noqa: F403
-from .ext import ExtType, Timestamp
+from .ext import Bypass, ExtType, Timestamp
 
 version = (1, 2, 1)
 __version__ = "1.2.1"
--- a/msgpack/_packer.pyx
+++ b/msgpack/_packer.pyx
@@ -8,7 +8,7 @@ from cpython.datetime cimport (
 cdef ExtType
 cdef Timestamp
 
-from .ext import ExtType, Timestamp
+from .ext import ExtType, Timestamp, Bypass
 
 
 cdef extern from "Python.h":
@@ -222,6 +222,8 @@ cdef class Packer:
             llval = o.seconds
             ulval = o.nanoseconds
             msgpack_pack_timestamp(&self.pk, llval, ulval)
+        elif type(o) is Bypass:
+            msgpack_pack_raw_body(&self.pk, o.data, len(o.data))
         elif PyList_CheckExact(o) if strict else (PyTuple_Check(o) or PyList_Check(o)):
             L = Py_SIZE(o)
             if L > ITEM_LIMIT:
--- a/msgpack/ext.py
+++ b/msgpack/ext.py
@@ -16,6 +16,19 @@ class ExtType(namedtuple("ExtType", "code data")):
         return super().__new__(cls, code, data)
 
 
+class Bypass:
+    """Bypass is a placeholder class to skip serialization and pass the bytes value as is."""
+
+    __slots__ = ["data"]
+
+    def __init__(self, data):
+        if isinstance(data, bytes):
+            self.data = data
+
+        else:
+            self.data = memoryview(data).tobytes()
+
+
 class Timestamp:
     """Timestamp represents the Timestamp extension type in msgpack.
 
--- a/msgpack/ext.py
+++ b/msgpack/ext.py
@@ -18,6 +18,19 @@ class ExtType(namedtuple("ExtType", "code data")):
         return super().__new__(cls, code, data)
 
 
+class Bypass:
+    """Bypass is a placeholder class to skip serialization and pass the bytes value as is."""
+
+    __slots__ = ["data"]
+
+    def __init__(self, data):
+        if isinstance(data, bytes):
+            self.data = data
+
+        else:
+            self.data = memoryview(data).tobytes()
+
+
 class Timestamp:
     """Timestamp represents the Timestamp extension type in msgpack.
 
--- a/msgpack/fallback.py
+++ b/msgpack/fallback.py
@@ -38,7 +38,7 @@ else:
 
 
 from .exceptions import BufferFull, ExtraData, FormatError, OutOfData, StackError
-from .ext import ExtType, Timestamp
+from .ext import Bypass, ExtType, Timestamp
 
 EX_SKIP = 0
 EX_CONSTRUCT = 1
@@ -772,6 +772,9 @@ class Packer:
                     self._buffer.write(struct.pack(">BI", 0xC9, L))
                 self._buffer.write(struct.pack("b", code))
                 self._buffer.write(data)
+                return
+            if check(obj, Bypass):
+                self._buffer.write(obj.data)
                 return
             if check(obj, list_types):
                 n = len(obj)
--- a/msgpack/fallback.py
+++ b/msgpack/fallback.py
@@ -38,7 +38,7 @@ else:
 
 
 from .exceptions import BufferFull, ExtraData, FormatError, OutOfData, StackError
-from .ext import ExtType, Timestamp
+from .ext import Bypass, ExtType, Timestamp
 
 EX_SKIP = 0
 EX_CONSTRUCT = 1
@@ -778,6 +778,9 @@ class Packer:
                     self._buffer.write(struct.pack(">BI", 0xC9, L))
                 self._buffer.write(struct.pack("b", code))
                 self._buffer.write(data)
+                return
+            if check(obj, Bypass):
+                self._buffer.write(obj.data)
                 return
             if check(obj, list_types):
                 n = len(obj)
--- a/test/test_bypass.py
+++ b/test/test_bypass.py
@@ -1,0 +1,57 @@
+from msgpack import Bypass, packb, unpackb
+
+
+def test_roundtrip_bypassed_bytes():
+    binary = b"\x00\x01\x02\x03\x04\x05"
+
+    binary_packed = packb(binary)
+    binary_bypassed_packed = packb(Bypass(binary_packed))
+
+    assert binary_bypassed_packed == binary_packed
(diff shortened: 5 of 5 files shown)


methane:
ja: JSONの中にJSONをネストするにはエスケープが必要になりますが、msgpackの中にmsgpackをネストして格納する場合はエスケープが必要ありません。
en: To nest JSON within JSON, escaping is required, but when nesting msgpack within msgpack, escaping is not necessary.

ja: そのため、このようなハックが有効になるようなデータを扱う場合は、msgpackをネストすることが良いプラクティスになります。
en: Therefore, when dealing with data that benefits from such a hack, nesting msgpack becomes a good practice.

ja: それでもこの機能が必要だと思うのであれば、同様の機能を実装している他のmsgpack実装を複数紹介してください。少なくとも1つはPython以外の言語用のライブラリを選んでください。
en: If you still believe this feature is necessary, please provide multiple examples of other msgpack implementations that have similar functionality. At least one of them should be a library for a language other than Python.

ja: もし他にこの機能を実装しているmsgpack実装が見つからず、それでもこの機能が本当に必要なんだと強く確信しているのであれば、その確信を私に伝えるための十分な根拠を提供してください。

en: If you cannot find any other msgpack implementations that have this feature, and you are still strongly convinced that this feature is truly necessary, please provide sufficient evidence to convince me of your conviction.

marcoffee:
Hi! Sorry for the time it took me to reply. Here are some use cases (and below I left some msgpack packages that do similar things):
- When you have an external class (or a class from a compiled module) that implements its own msgpack serialization, it is useful (and much more efficient) to allow the class to create its own msgpack and then concatenate it into the main output (so we just bypass the class output into the main payload). That way, we avoid reimplementing the serialization logic from the class (which is error-prone: two places to update every time).
- When you receive a msgpack payload (let's say on an API) and want to just pass it along to the next service, `Bypass` is useful to avoid having to deserialize and re-serialize the payload (just place it as a msgpack object inside the main payload, and it's done).
- When serializing the same object multiple times (especially big objects), `Bypass` allows for a cache where you serialize it once and then reuse the serialized payload on the next iterations.

Note that just placing the serialized msgpack as a `bin` (or using an extension type for that) also increases:
1. The cost of serialization (a little bit, you need to serialize the binary header / size).
2. The size of the payload (a little bit per binary, since you need to include its header / size, but it grows with the number of embedded binaries, so it can become significant).
3. Mostly, the cost (and implementation logic) of de-serialization: you need to deserialize the main payload first, then access the object's raw binary and deserialize it, which could even be recursive, so it is very error-prone.

The `Bypass` class allows for flattening this object payload into the main payload, in a fast and safe (if you know it is valid msgpack) way. The resulting payload is byte-for-byte identical to serializing the object inline, so `Bypass` is fully transparent to the decoding side (any language / any msgpack library, it does not even need to know `Bypass` was used), while the raw binary (or extension type) changes the output format and forces every consumer to know the convention (possibly recursive if there are many steps).

I resorted to `Bypass` here because the current implementation does not provide a stream where you can write freely, but some of the example packages below do provide a stream, so it is a way to avoid the `Bypass` class (and another option if you want to go that way, but that would change the current `default` param or introduce a new one, e.g., `streamed_default` or something similar, which would increase implementation complexity / time).

Concatenating `packb` outputs externally does work, but one would need to assemble every enclosing container by hand (`pack_array_header` / `pack_map_header`, then concatenate, which could turn into O(n^2) complexity if done wrong): a `Bypass` value can be placed anywhere inside a Python structure (deep inside dicts and/or lists passed to `packb`), which manual concatenation cannot achieve without reimplementing the packer's container logic. Here are some examples of msgpack packages outside Python that allow doing something similar:
- [vmihailenco/msgpack](https://github.com/vmihailenco/msgpack) (language: Go) - has a `RawMessage` type that allows for bypassing the serialization of a msgpack object ([here](https://github.com/vmihailenco/msgpack/blob/19c91dfdfa062658c39d9321be26163fc5833bd1/msgpack.go#L23-L32)), which is written as-is to the stream by the internal `Encoder.write` method ([here](https://github.com/vmihailenco/msgpack/blob/19c91dfdfa062658c39d9321be26163fc5833bd1/encode.go#L262)). It also exposes the underlying stream through `Encoder.Writer` ([here](https://github.com/vmihailenco/msgpack/blob/19c91dfdfa062658c39d9321be26163fc5833bd1/encode.go#L96)), so you can also write to it directly.
- [msgpack/msgpack-java](https://github.com/msgpack/msgpack-java) (language: Java) - has a `MessagePacker.writePayload` method that allows for writing a msgpack object directly to the stream ([here](https://github.com/msgpack/msgpack-java/blob/6d26864835f66fa34160c19ccf90924856091eb9/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java#L1095)), and a zero-copy `addPayload` variant ([here](https://github.com/msgpack/msgpack-java/blob/6d26864835f66fa34160c19ccf90924856091eb9/msgpack-core/src/main/java/org/msgpack/core/MessagePacker.java#L1133)). Note that, in this case, their docs mention they are "used with" `packRawStringHeader` / `packBinaryHeader`, but that is descriptive, not a requirement: both are plain verbatim copies to the output, and their names point more to "append to the buffer" than to "pack the body of the started string/binary".
- [msgpack/msgpack-c](https://github.com/msgpack/msgpack-c) (language: [C](https://github.com/msgpack/msgpack-c/tree/c_master) and [C++](https://github.com/msgpack/msgpack-c/tree/cpp_master)) - you own the stream (it only stores a [reference in C++](https://github.com/msgpack/msgpack-c/blob/a5250ecbcae1d34346c7445345774b0abbd651bc/include/msgpack/v1/pack.hpp#L633) or a [pointer in C](https://github.com/msgpack/msgpack-c/blob/1be50ee5dadc4b85f79945c51df362582ccb2631/include/msgpack/pack.h#L132) to it), so you can write your objects directly to the stream when, for example, creating an array or a map. That one requires the container-header "solution" mentioned above (via `pack_array` / `pack_map`), but it is still a valid example of bypassing the packer logic and writing directly to the stream. They also allow for the `msgpack-java` workaround: [msgpack_pack_str_body](https://github.com/msgpack/msgpack-c/blob/1be50ee5dadc4b85f79945c51df362582ccb2631/include/msgpack/pack.h#L91) / [msgpack_pack_v4raw_body](https://github.com/msgpack/msgpack-c/blob/1be50ee5dadc4b85f79945c51df362582ccb2631/include/msgpack/pack.h#L95) / [msgpack_pack_bin_body](https://github.com/msgpack/msgpack-c/blob/1be50ee5dadc4b85f79945c51df362582ccb2631/include/msgpack/pack.h#L98) (in C) and [packer::pack_str_body](https://github.com/msgpack/msgpack-c/blob/a5250ecbcae1d34346c7445345774b0abbd651bc/include/msgpack/v1/pack.hpp#L513-L523) / [packer::pack_v4raw_body](https://github.com/msgpack/msgpack-c/blob/a5250ecbcae1d34346c7445345774b0abbd651bc/include/msgpack/v1/pack.hpp#L539-L550) / [packer::pack_bin_body](https://github.com/msgpack/msgpack-c/blob/a5250ecbcae1d34346c7445345774b0abbd651bc/include/msgpack/v1/pack.hpp#L565-L575) (in C++) allow for writing a msgpack object directly to the stream, but in this case their names are more explicit about what they should be used for.
- [msgpack/msgpack-ruby](https://github.com/msgpack/msgpack-ruby) (language: Ruby) - the packer's buffer is public (accessible through `Packer#buffer`), and its [`MessagePack::Buffer#write`](https://ruby.msgpack.org/MessagePack/Buffer.html#write-instance_method) method appends the given bytes verbatim, with no type header ([here, implemented in C](https://github.com/msgpack/msgpack-ruby/blob/b136b7fc114ed5935f0492dc31c0eae52613f263/ext/msgpack/buffer_class.c#L278-L288)).

#### Extra cases:

##### Other msgpack packages in Python

Inside Python, [msgspec](https://github.com/msgspec/msgspec) (which also implements msgpack) has a `Raw` type that does exactly this (from its [docs](https://msgspec.dev/api): "Raw objects wrap pre-encoded messages. These can be added as components of larger messages without having to pay the cost of decoding and re-encoding them"). Note that `Raw` also appears on the decoding side, but only as a type annotation: a field annotated as `Raw` is kept undecoded (as a view into the original message, to be decoded later by the user); regular decoding does not produce `Raw` objects from `Raw`-encoded instances.

##### On CBOR

The same passthrough also exists in CBOR (a binary format very similar to msgpack), for example:
- [fxamacker/cbor](https://github.com/fxamacker/cbor) (language: Go) - has a `RawMessage` type that is written to the output with no CBOR framing added ([here](https://pkg.go.dev/github.com/fxamacker/cbor/v2#RawMessage)).
- [cbor2](https://github.com/agronholm/cbor2) (language: Python) - has a `CBOREncoder.write` method that copies the given bytes directly to the output ([here](https://cbor2.readthedocs.io/en/stable/api.html)).
- [QCBOR](https://github.com/laurencelundblade/QCBOR) (language: C) - has a `QCBOREncode_AddEncoded` function ("Add some already-encoded CBOR bytes") ([here](https://github.com/laurencelundblade/QCBOR/blob/148a71014146a004043b599f990ff2b78b488e1a/inc/qcbor/qcbor_encode.h#L1964)).
- [System.Formats.Cbor](https://learn.microsoft.com/en-us/dotnet/api/system.formats.cbor.cborwriter.writeencodedvalue?view=net-8.0) (language: C#/.NET) - has a `CborWriter.WriteEncodedValue` method that validates the encoded value and copies it verbatim to the output.

About the name: `Raw` would follow msgspec's `Raw` and Go's / fxamacker's `RawMessage`, but it could be confused with the `bin` format, so I avoided it. Maybe a better name for the `Bypass` class would be `Passthrough` or `Flatten`, but I am not sure if any of them is better. I am always open to suggestions.

Note: I have updated my branch of the repository to include the latest commits and add some basic `Bypass` tests.

PR #664, opened 2026-02-20
ci: add riscv64 manylinux/musllinux wheels

justeph:
Now that cibuildwheel and PyPI support riscv64, we can start building riscv64 wheels.

Because there is no native riscv64 runner available, this PR adds a QEMU-based riscv64 job to the cibuildwheel workflow.

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -12,9 +12,19 @@ jobs:
     strategy:
       matrix:
         # macos-13 is for intel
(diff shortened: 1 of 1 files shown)


justeph:
FYI, [RISE](https://riseproject.dev/) has been building and distributing msgpack manylinux wheels for riscv64 for several months (versions 1.1.2, 1.1.1, 1.1.0 and 1.0.8). Tested builds are available for riscv64 [here](https://riseproject.gitlab.io/python/wheel_builder/packages/msgpack.html).

justeph:
Here is the run on my local fork https://github.com/justeph/msgpack-python/actions/runs/22229666154/job/64305563827

PR #665, opened 2026-04-18
fix: check return code in `unpack_callback_int64`

KowalskiThomas:
This simply adds a null pointer check after calling `PyLong_FromLongLong` / `PyLong_FromLong` like other similar functions do it ([example here](https://github.com/msgpack/msgpack-python/blob/5142b3d1b1929104103a14c1d7e148b8b8bb7896/msgpack/unpack.h#L71-L83)).

### DIFF
--- a/msgpack/unpack.h
+++ b/msgpack/unpack.h
@@ -103,7 +103,7 @@ static inline int unpack_callback_int64(unpack_user* u, int64_t d, msgpack_unpac
         p = PyLong_FromLongLong((PY_LONG_LONG)d);
     } else {
         p = PyLong_FromLong((long)d);
-    }
+    PyObject *p = PyLong_FromLongLong((PY_LONG_LONG)d);
     if (!p)
         return -1;
     *o = p;
--- a/msgpack/unpack.h
+++ b/msgpack/unpack.h
@@ -70,12 +70,7 @@ static inline int unpack_callback_uint32(unpack_user* u, uint32_t d, msgpack_unp
 
 static inline int unpack_callback_uint64(unpack_user* u, uint64_t d, msgpack_unpack_object* o)
 {
-    PyObject *p;
-    if (d > LONG_MAX) {
-        p = PyLong_FromUnsignedLongLong((unsigned PY_LONG_LONG)d);
-    } else {
-        p = PyLong_FromLong((long)d);
-    }
+    PyObject *p = PyLong_FromUnsignedLongLong((unsigned PY_LONG_LONG)d);
     if (!p)
         return -1;
     *o = p;
--- a/msgpack/unpack.h
+++ b/msgpack/unpack.h
@@ -98,11 +98,6 @@ static inline int unpack_callback_int8(unpack_user* u, int8_t d, msgpack_unpack_
 
 static inline int unpack_callback_int64(unpack_user* u, int64_t d, msgpack_unpack_object* o)
 {
-    PyObject *p;
-    if (d > LONG_MAX || d < LONG_MIN) {
-        p = PyLong_FromLongLong((PY_LONG_LONG)d);
-    } else {
-        p = PyLong_FromLong((long)d);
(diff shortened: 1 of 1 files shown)


msgpack/unpack.h
ThomasWaldmann: Should this be `PyLong_FromUnsignedLong((unsigned long)d)`?

ThomasWaldmann: That "|| d < LONG_MIN" looks strange, is it correct?

methane: Both are correct. Since we checked `d <= LONG_MAX`, `(long)d` is guaranteed to fit in a signed long.
But `p = PyLong_FromUnsignedLongLong((unsigned PY_LONG_LONG)d)` without `d > LONG_MAX` check would be better.

ThomasWaldmann: OK, agreed.

But if both are correct, one could use the less eyebrows-raising  "unsigned" call and cast?

KowalskiThomas: I do believe it's correct -- long could be less than 64 bits in which case you want to make sure either the number is positive and less than the max a long can fit, or it's negative and it's more than the smallest negative number a long can fit. 

Am I missing something or does that clear things up?

ThomasWaldmann: Clear now, thanks.

KowalskiThomas: Do you want me to change this in the PR?

ThomasWaldmann: @methane decides.

It is somehow unrelated to the PR topic, I just stumbled over it while looking at the code.

methane: > But if both are correct, one could use the less eyebrows-raising "unsigned" call and cast?

There is no `PyLong_FromUnsignedLong()`. `PyLong_FromLong` requires `(unsigned long)` cast to suppress implicit cast warning.


> Do you want me to change this in the PR?

Please remove `if (d > LONG_MAX) {` and use `PyLong_FromUnsignedLongLong` always.

ThomasWaldmann: > There is no PyLong_FromUnsignedLong()

There is a call to that in line 333.

methane: You are right. But using only `PyLong_FromUnsignedLongLong` is much simpler.

KowalskiThomas: I removed the call and now it's always `PyLong_FromUnsignedLongLong`!

msgpack/unpack.h
```diff
@@ -70,12 +70,7 @@ static inline int unpack_callback_uint32(unpack_user* u, uint32_t d, msgpack_unp
 
 static inline int unpack_callback_uint64(unpack_user* u, uint64_t d, msgpack_unpack_object* o)
 {
-    PyObject *p;
-    if (d > LONG_MAX) {
-        p = PyLong_FromUnsignedLongLong((unsigned PY_LONG_LONG)d);
-    } else {
-        p = PyLong_FromLong((long)d);
-    }
+    PyObject *p = PyLong_FromUnsignedLongLong((unsigned PY_LONG_LONG)d);
```
ThomasWaldmann, line 73: Interesting, but it seems correct.

(I am not really familiar with the C api.)

msgpack/unpack.h
```diff
@@ -109,6 +104,8 @@ static inline int unpack_callback_int64(unpack_user* u, int64_t d, msgpack_unpac
     } else {
         p = PyLong_FromLong((long)d);
     }
```
ThomasWaldmann: but if that other change is correct, then this one can also be simplified in the same way, right?

methane: yes

methane (approved):

msgpack/unpack.h
methane:
```suggestion
    PyObject *p = PyLong_FromLongLong((PY_LONG_LONG)d);
```

methane (approved):

PR #667, opened 2026-04-21
fix: re-raise existing exception when available

KowalskiThomas:
## What is this PR?

This PR addresses an issue I found while working on #666 and reported in the description there. 

Looking at the usages of `unpack_callback_uint32` ([here for example](https://github.com/msgpack/msgpack-python/blob/6357bc272c83fb450fbe7de7e7ca15d815e75174/msgpack/_unpacker.pyx#L516-L523), which calls into [this](https://github.com/msgpack/msgpack-python/blob/6357bc272c83fb450fbe7de7e7ca15d815e75174/msgpack/_unpacker.pyx#L455-L485)) it seems like it checks for error values -2 and -3, but -1 will just `raise ValueError("Unpack failed: error = %d" % (ret,))`.

It believe it should just `raise` to raise the existing exception set by `PyErr_SetString` if it exists. This PR addresses that problem here and another place.

### DIFF
--- a/msgpack/_unpacker.pyx
+++ b/msgpack/_unpacker.pyx
@@ -205,7 +205,10 @@ def unpackb(object packed, *, object object_hook=None, object list_hook=None,
         raise FormatError
     elif ret == -3:
         raise StackError
-    raise ValueError("Unpack failed: error = %d" % (ret,))
+    elif PyErr_Occurred():
+        raise
+    else:
+        raise ValueError("Unpack failed: error = %d" % (ret,))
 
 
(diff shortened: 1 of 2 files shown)


methane:
Would it be easy to add a test case for this issue?

KowalskiThomas:
Ah yeah good point, let me have a look :)

KowalskiThomas:
I added a test case, but I don't think it's possible to easily cover/simulate something that would happen like in #666. But the tests do flex the `PyErr_Occurred` path.

PR #672, opened 2026-05-11
fix: properly handle return codes in `pack_timestamp`

KowalskiThomas:
## What is this PR?

This PR makes it so that return codes of helpers called in `msgpack_pack_ext` are properly checked _and surfaced if they're indicative of a failure_. 

**Note** the "big" diff at the end of the function is an indentation change on top of the actual change -- the code used to be indented with three spaces instead of four. I changed that (should be clear looking at the [diff without whitespace changes](https://github.com/msgpack/msgpack-python/pull/672/changes?w=1)).

### DIFF
--- a/msgpack/pack_template.h
+++ b/msgpack/pack_template.h
@@ -551,6 +551,7 @@ static inline int msgpack_pack_ext(msgpack_packer* x, char typecode, size_t l)
  */
 static inline int msgpack_pack_timestamp(msgpack_packer* x, int64_t seconds, uint32_t nanoseconds)
 {
+    int ret;
     if ((seconds >> 34) == 0) {
         /* seconds is unsigned and fits in 34 bits */
         uint64_t data64 = ((uint64_t)nanoseconds << 34) | (uint64_t)seconds;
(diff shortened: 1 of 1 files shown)


msgpack/pack_template.h
```diff
@@ -551,33 +551,41 @@ static inline int msgpack_pack_ext(msgpack_packer* x, char typecode, size_t l)
...
+        _msgpack_store32(&buf[0], nanoseconds);
+        _msgpack_store64(&buf[4], seconds);
+        ret = msgpack_pack_ext(x, -1, 12);
+        if (ret != 0)
+            return ret;
+
+        return msgpack_pack_raw_body(x, buf, 12);
     }
-    return 0;
 }
```
KowalskiThomas, line 589: If this was my code, I would probably prefer doing something like

```cpp
if (condition) {
  // whatever
  return something;
}

// the contents of the else
```

to make it clear to the reader that _we never not return a value from the function_. This is even more relevant now that I've removed the `return 0` at the end. But it's not my code so... let me know.

methane (approved):

Issue #674, opened 2026-05-17
msgpack.fallback DEFAULT_RECURSE_LIMIT smaller than python 3.10 default


rsiemens:
Hi there. I was running the pure python `msgpack.fallback` implementation against this [sample.json](https://github.com/Tencent/rapidjson/blob/master/bin/data/sample.json) test file in the tencent/rapidjson repo. Packing hit a `ValueError: recursion limit exceeded`.

It looks like this stems from the [DEFAULT_RECURSE_LIMIT](https://github.com/msgpack/msgpack-python/blob/main/msgpack/fallback.py#L55). Changing to use the `sys.getrecursionlimit()` resolves as it has a [default of 1000](https://github.com/python/cpython/blob/3.10/Python/ceval.c#L799) in 3.10.  Looking at the pyproject.toml it seems like support is >= python 3.10 so I think it should be fine to update the constant, or drop it completely in favor of just inheriting the python stack limit. An advantage of doing that is the users ability to control it with `sys.setrecursionlimit()`. Also then a `RecursionError` can be raised instead of `ValueError`.

Simple reproduction:

```python
from msgpack.fallback import Packer
import json

with open('sample.json') as f:
    Packer().pack(json.load(f))  # ValueError
```

This works fine, but the public `Packer.pack` api doesn't expose it directly:

```python
from msgpack.fallback import Packer
import json
import sys

with open('sample.json') as f:
    Packer()._pack(json.load(f), nest_limit=sys.getrecursionlimit())  # 1000
```

Thanks!

methane:
I don't remember why I added nest_limit and why I chose 511 for it.
It might be related to Python's `sys.getrecursionlimit()`, but I don't think that was the only reason.
The fallback was implemented after Cython version, and nest_limit has been around since Cython version.

My guess is that it was likely a balance between data compatibility and convenience. The C implementation of msgpack originally only supported 32 levels of nesting. Since that was too strict, the Cython version expanded it to 1024 levels for unpacking, while limiting packing to 512-1 levels. I don't remember why the same number wasn't used, but perhaps it was considering the [Robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) "be conservative in what you do, be liberal in what you accept from others".

In any case, I am opposed to changes that would widen the difference in behavior between the Cython version and the fallback.
Also, I am opposed to generating data with deep nesting that would cause the Unpacker in the Cython version to fail to unpack.

[...]

PR #675, opened 2026-05-17
issue 674: remove DEFAULT_RECURSE_LIMIT and ValueError for signaling recursion limit

rsiemens:
If issue [674](https://github.com/msgpack/msgpack-python/issues/674) is accepted then this patch resolves.

If removing `nest_limit` is deemed to risky (fear of users depending on this) then can keep it and update `DEFAULT_RECURSE_LIMIT = sys.getrecursionlimit()`.


rsiemens:
resolved by https://github.com/msgpack/msgpack-python/pull/676

PR #682, opened 2026-06-02
Bump sys.setrecursionlimit within test_nest_limit_1024

clin1234:
Fixes test failures with no-GIL interpreters


test/test_limits.py
methane, line 166: Why was 10000 not enough? The nest is only 1024....

methane:
It didn't fix the issue.
https://github.com/msgpack/msgpack-python/actions/runs/26861967262/job/79217458032?pr=685

PR #718, opened 2026-07-21
Allow default handler to take a position parameter

PeterC-DLS:
This enables padding to be computed when packing a custom type and enables zero-copy unpacking by aligning primitives correctly. The code in `test_default` demonstrates correct padding for `array`s.

This is motivated by passing NumPy's `ndarray`s with their raw data in `bin` format to a web frontend where `TypedArray`s must be correctly aligned otherwise a copy must be taken by slicing the underlying `ArrayBuffer` (in the `@msgpack/msgpack` package). Note, the Javascript library gives access to the encoder context for extension types but this is not available for this Python library.

### DIFF
--- a/msgpack/__init__.py
+++ b/msgpack/__init__.py
@@ -4,8 +4,8 @@ import os
 from .exceptions import *  # noqa: F403
 from .ext import ExtType, Timestamp
 
-version = (1, 2, 1)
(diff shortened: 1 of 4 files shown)


methane:
memo: https://github.com/msgpack/msgpack-javascript/pull/248

PR #728, opened 2026-08-26
Raise OverflowError when use_single_float cannot represent a value

dylanpulver:
Fuzzing the Cython packer against the pure Python fallback over 4000 random objects, `packb` disagreed on 38 of them. Every disagreement was the same case: `use_single_float=True` with a value above FLT_MAX.

```
# C extension
>>> msgpack.packb(1e300, use_single_float=True).hex()
'ca7f800000'          # +inf

# MSGPACK_PUREPYTHON=1
>>> msgpack.packb(1e300, use_single_float=True)
OverflowError: float too large to pack with f format
```

Two things kept the C path quiet. `_packer.pyx` cast the value with `<float>o` before the call, so `msgpack_pack_float` only ever saw a narrowed float. And `msgpack_pack_float` discarded the return value of `PyFloat_Pack4`, which is what reports the overflow.

Passing the double through and checking that return value covers both. `PyFloat_Pack4` backs `struct.pack('>f', ...)`, so the exception type and message now match the fallback exactly.

Out-of-range integers are handled this way today: `packb(2**64)` raises `PackOverflowError` from either packer, pinned by `test_limits.py::test_integer`. The new test sits next to it.

Reverting the source makes the new test fail on the C extension. It passes either way under `MSGPACK_PUREPYTHON=1`, where it pins existing behaviour. After the change the same fuzz run reports 0 of 4000 disagreements on `packb`.

Infinity is representable in single precision and packs unchanged, which the test also covers.

### DIFF
--- a/msgpack/_packer.pyx
+++ b/msgpack/_packer.pyx
@@ -27,7 +27,7 @@ cdef extern from "pack.h":
     int msgpack_pack_false(msgpack_packer* pk) except -1
     int msgpack_pack_long_long(msgpack_packer* pk, long long d) except -1
     int msgpack_pack_unsigned_long_long(msgpack_packer* pk, unsigned long long d) except -1
-    int msgpack_pack_float(msgpack_packer* pk, float d) except -1
+    int msgpack_pack_float(msgpack_packer* pk, double d) except -1
     int msgpack_pack_double(msgpack_packer* pk, double d) except -1
     int msgpack_pack_array(msgpack_packer* pk, size_t l) except -1
     int msgpack_pack_map(msgpack_packer* pk, size_t l) except -1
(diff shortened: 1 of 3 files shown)


dylanpulver:
The one red job (`3.15t on windows-latest`) does not come from this change.

Its step `Test (C extension) in parallel under free-threading` passed, and that is the path this PR touches. The job then spent 46 minutes inside `Test (pure Python fallback) in parallel under free-threading` and was killed there.

That stall reproduces on `e94e1dc` with no patch applied. `MSGPACK_PUREPYTHON=1 pytest --parallel-threads=auto --iterations=20 test` parks on `test_case.py::test_array16` on an 18 core machine. At `--parallel-threads=4` the same suite finishes in about 25 seconds on both `e94e1dc` and this branch. `auto` scales with runner CPU count, so the cost of `test_array16` and `test_array32` under the pure Python packer grows with the size of the runner.

A re-run should come back green.

